PR #30, opened 2015-10-28
Codepush sync.

dlebu:
Sync is a new, one line operation that makes apps' integration with CodePush easier.

@lostintangent @nisheetjain @mslaguana @silhouettes @shishirx34 @dtivel


www/syncStatus.ts
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * An update is available, it has been downloaded, unzipped and copied to the deployment folder.
+     * After the completion of the callback invoked with SyncStatus.APPLY_SUCCESS, the application will be reloaded with the updated code and resources.
+     */
+    APPLY_SUCCESS,
+    
+    /**
+     * An optional update is available, but the user declined to install it. The update was not downloaded.
+     */
+    USER_DECLINED,
```
lostintangent: I called this state "UPDATE_IGNORED" on the React Native side, but I'm open to either one

www/syncStatus.ts
```diff
@@ -0,0 +1,32 @@
...
+"use strict";
+
+/**
+ * Defines the possible result statuses of the window.codePush.sync operation.
+ */
+enum SyncStatus {
+    /**
+     * The application is up to date.
+     */
+    UP_TO_DATE,
```
lostintangent: I called this state "NO_UPDATE_AVAILABLE" on the RN side, but I'm not married to either one :)

www/codePush.ts
```diff
@@ -108,6 +115,119 @@ class CodePush implements CodePushCordovaPlugin {
...
+     */
+    private getDefaultSyncOptions(): SyncOptions {
+        if (!CodePush.DefaultSyncOptions) {
+            CodePush.DefaultSyncOptions = {
+                dialogTitle: "Update",
+                mandatoryUpdateMessage: "You will be updated to the latest version.",
+                mandatoryUpdateContinueButtonLabel: "Continue",
+                optionalUpdateMessage: "An update is available. Would you like to install it?",
+                optionalUpdateConfirmButtonLabel: "Install",
+                optionalUpdateCancelButtonLabel: "Cancel",
```
lostintangent: I was calling this "IgnoreButton" on the RN side, with the default text being "Ignore". I'm cool either way

plugin.xml
```diff
@@ -23,6 +24,10 @@
             <clobbers target="RemotePackage" />
         </js-module>
 
+        <js-module src="bin/www/syncStatus.js" name="syncStatus">
+            <clobbers target="SyncStatus" />
```
lostintangent: Is it idiomatic on the Cordova side for this to be a global enum? On the RN side, I made the SyncStatus enum hang off the CodePush module itself, since RN relies on modules instead of global vars

lostintangent:
LGTM! I'm excited to get this in :)

www/syncStatus.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference path="../typings/codePush.d.ts" />
+
+"use strict";
+
+/**
+ * Defines the possible result statuses of the window.codePush.sync operation.
+ */
+enum SyncStatus {
```
silhouettes: Would it make sense to put this enum in the common SDK? There's a lot of overlap here. Although, there is a small inconsistency in that the Cordova implementation exposes an ERROR enum, whereas the RN implementation rejects the promise.

lostintangent: That makes sense :)

lostintangent:
I ended up adding another sync option to the RN plugin called "useReleaseDescription" that allows the dev to specify whether they want the alert dialog's message to be derived from the remote package's description, as opposed to using the two different option fields to define a static string for optional and mandatory updates. That way, the sync implementation respects all meaningful metadata attributes of a release (e.g. mandatory, failedApply, description) in order to allow the dev to drive their experience from the CLI/server. What do you think?

lostintangent:
If you could update to sample app to use sync that would be great, that way we lead with the simple solution first. Otherwise, this LGTM!

www/syncStatus.ts
dlebu: I renamed it to UPDATE_IGNORED for consistency.

www/syncStatus.ts
dlebu: I think UP_TO_DATE is better since it is positive.

www/codePush.ts
dlebu: I changed it to "Ignore".

plugin.xml
dlebu: Yes it's pretty common to register global types in plugins. We did the same with LocalPackage and RemotePackage types.

www/codePush.ts
```diff
@@ -105,9 +112,122 @@ class CodePush implements CodePushCordovaPlugin {
...
+        };
+
+        var onUpdate = (remotePackage: RemotePackage) => {
+            if (!remotePackage || (remotePackage.failedApply && syncOptions.ignoreFailedUpdates)) {
+                syncCallback && syncCallback(SyncStatus.UP_TO_DATE);
+            } else {
+                if (remotePackage.isMandatory && syncOptions.mandatoryUpdateMessage) {
+                    /* Alert user */
+                    navigator.notification.alert(syncOptions.mandatoryUpdateMessage, () => { downloadAndInstallUpdate(remotePackage); }, syncOptions.dialogTitle, syncOptions.mandatoryUpdateContinueButtonLabel);
+                } else if (!remotePackage.isMandatory && syncOptions && syncOptions.optionalUpdateMessage) {
```
silhouettes: Nittish, but you don't need to check syncOptions exists here (and aren't doing that in the 'if' above anyway)

silhouettes:
(Looks good!)

www/codePush.ts
dlebu: You are right. That was written before I introduced the default options.

dlebu:
@lostintangent I think that is a good idea. Per our discussion this morning, we are going to add an appendReleaseDescription flag and a descriptionPrefix flag, which control the way the description appears in the alert message.

lostintangent:
I looked at all your latest changes, and they look great. Ship this baby!

PR #32, opened 2015-11-02
Fix for iOS 9 NSUserDefaults immutable change.

dlebu:
Fixing a bug found by @lostintangent where on iOS 9 NSUserDefaults is immutable once created.
@dtivel @silhouettes @geof90

### DIFF
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -311,13 +311,9 @@ NSString* const CurrentPackageManifestName = @"currentPackage.json";
 
 - (void)saveFailedUpdate:(NSString *)packageHash {
     NSUserDefaults *preferences = [NSUserDefaults standardUserDefaults];
-    NSMutableArray* failedUpdates = [preferences objectForKey:FailedUpdatesKey];
+    NSMutableArray* failedUpdates = [[preferences objectForKey:FailedUpdatesKey] mutableCopy];
     if (nil == failedUpdates) {
         failedUpdates = [[NSMutableArray alloc] init];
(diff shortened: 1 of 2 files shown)


dtivel:
looks good

src/ios/CodePush.m
```diff
@@ -314,6 +314,10 @@ - (void)saveFailedUpdate:(NSString *)packageHash {
     NSMutableArray* failedUpdates = [preferences objectForKey:FailedUpdatesKey];
```
geof90: Since this **not** actually an NSMutableArray, just for added clarity, should we store [preferences objectForKey:FailedUpdatesKey] into a normal immutable NSArray variable first, then call ```mutableCopy``` on it?

geof90:
LGTM

src/ios/CodePush.m
dlebu: Cool. I modified the implementation to call `mutableCopy` right after we get the object. Hopefully it looks more clear now.

geof90: if ```[preferences objectForKey:FailedUpdatesKey]``` is a ```nil```, calling ```mutableCopy``` on it right after wouldn't work right?

dlebu: It works fine, I triple checked it. `mutableCopy` works on any `NSObject`, and for `nil` it just returns `nil` again.

geof90: Okay. I was being paranoid then. This looks good to me. :)

PR #33, opened 2015-11-03
Bumped version to 1.1.0-beta.

dlebu:
I did not bump the dependency version of code-push, since in the latest release the [plugin.xml version was not increased](https://github.com/Microsoft/code-push/blob/master/sdk/plugin.xml#L2). This is okay for now since the acquisition code was not changed. We will bump all versions in the next release.

@lostintangent @geof90 @dtivel @silhouettes @nisheetjain

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cordova-plugin-code-push",
-  "version": "1.0.1-beta",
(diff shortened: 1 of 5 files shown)


lostintangent:
LGTM!

Issue #39, opened 2015-11-11
CodePush-ed apps always navigate to the start page when restarted [Android]


geof90:
Repro Steps:

1. Install CodePush app update 
2. Navigate to a different page in the app
3. Press the home button (send app to background)
4. Re-enter the app

Result:

Web view does not re-open in the current page, but navigates back to the start page due to https://github.com/Microsoft/cordova-plugin-code-push/blob/master/src/android/CodePush.java#L324 

This behaviour is unique to CodePush-ed apps. A normal cordova app will re-open in the current page before being sent to the background. Also, the behavior only appears for apps that have already been updated once or more via CodePush, it does not occur If the app is a fresh install (from Play Store).

geof90:
Closed with https://github.com/Microsoft/cordova-plugin-code-push/pull/40

PR #40, opened 2015-11-15
only call app start once

ojotoxy:
The obvious way of fixing this issue: https://github.com/Microsoft/cordova-plugin-code-push/issues/39


dlebu:
Thanks for the fix @ojotoxy! The code looks good to me. @geof90, can you check if this commit fixes your issue?

geof90:
I have verified that this fix works :+1:

PR #41, opened 2015-11-15
Clarify usage of syncOptions

ojotoxy:


lostintangent:
Thanks for adding this clarification! That said, I'm not sure this behavior is expected :) The React Native plugin allows you to overwrite any of the individual default values, or specify none at all, while still "inheriting" the non-overwritten defaults, so if that isn't the case with the Cordova plugin, I'd be interested to discuss us fixing that. What are your thoughts?

ojotoxy:
I agree it should probably be changed. It was pretty surprising to have it not inherit from the defaults.

dlebu:
I definitely agree as well we should change the behavior if it is unexpected. The update messages and labels are the only ones that are [not set to the default values](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/codePush.ts#L146) if they are not specified. That is the way we disable the dialog right now. If we make them inherit from defaults, the users would have to set the messages to null to disable the dialog. Would that make more sense or do you think we need to introduce another boolean option here to control the UI?

dlebu:
@ojotoxy We just released version [1.3.0-beta](https://github.com/Microsoft/cordova-plugin-code-push/releases/tag/1.3.0-beta) to GitHub and npm. This fixed the default options issue you reported. Please note that the default `SyncOptions` have changed. With the default options the updates are installed on the next restart of the application (this is a new feature) and no user dialog is shown. Please try it out and let us know if you have more feedback.

PR #53, opened 2016-01-14
1.5.0-beta

dlebu:
This PR contains:
 - Breaking changes to the rollback mechanism. Rollback is now enabled by default and the rollbackTimeout was removed. `codePush.notifyApplicationReady()` needs to be called after update, sometime during the first run of the updated code

- Added a two new methods: `getPendingPackage()` and `restartApplication()`
- Changes to the reporting mechanism, to accomodate statistics in the CodePush CLI

The PR will _not_ be merged until the new acquisition bits are published. I assumed that the new `code-push` (acquisition SDK) version will be `1.5.0-beta`.

@lostintangent @dtivel @geof90 @silhouettes @shishirx34 @ryuyu @nisheetjain


bin/www/codePush.js
```diff
@@ -26,8 +26,28 @@ var CodePush = (function () {
     CodePush.prototype.notifyApplicationReady = function (notifySucceeded, notifyFailed) {
         cordova.exec(notifySucceeded, notifyFailed, "CodePush", "updateSuccess", []);
     };
+    CodePush.prototype.restartApplication = function (installSuccess, errorCallback) {
```
lostintangent: I'd prefer `restartApp`, since "app" is commonly used for mobile apps, whereas "application" is more for desktop apps (the actual definitions of these aren't quite as simple). This is what we went with in the React plugin, so if you feel strongly about it, let's chat :)

bin/www/localPackage.js
```diff
@@ -91,11 +91,16 @@ var LocalPackage = (function (_super) {
...
+                                installSuccess && installSuccess();
+                                cordova.exec(function () { }, function () { }, "CodePush", "install", [deployDir.fullPath, installOptions.installMode.toString()]);
+                            }
+                            else {
+                                cordova.exec(function () { installSuccess && installSuccess(); }, function () { installError && installError(); }, "CodePush", "install", [deployDir.fullPath, installOptions.installMode.toString()]);
+                            }
                         };
                         var preInstallSuccess = function () {
-                            Sdk.reportStatus(AcquisitionStatus.DeploymentSucceeded);
+                            Sdk.reportStatus(AcquisitionStatus.DeploymentSucceeded, _this);
```
geof90: `reportStatus()` is now used to record that a user has either 1. upgraded successfully to a new version or 2. failed to upgrade to a new version (due to a crash triggering a rollback, or an install failure). Therefore, I think we should report DeploymentSucceeded during `notifyApplicationReady()` if we see that the currentPackage has isFirstRun set to true, because that is the only time we can ascertain that an app has been installed successfully.

Also, we will need a new mechanism to find out that an app has been rolled back, and call reportStatus() there.

www/localPackage.ts
```diff
@@ -118,13 +117,18 @@ class LocalPackage extends Package implements ILocalPackage {
...
+                                /* no neeed for callbacks, the javascript context will reload */
+                                cordova.exec(() => { }, () => { }, "CodePush", "install", [deployDir.fullPath, installOptions.installMode.toString()]);
+                            } else {
+                                cordova.exec(() => { installSuccess && installSuccess(); }, () => { installError && installError(); }, "CodePush", "install", [deployDir.fullPath, installOptions.installMode.toString()]);
+                            }
                         };
 
                         var preInstallSuccess = () => {
-                            Sdk.reportStatus(AcquisitionStatus.DeploymentSucceeded);
+                            Sdk.reportStatus(AcquisitionStatus.DeploymentSucceeded, this);
```
geof90: `reportStatus()` is now used to record that a user has either:
1. upgraded successfully to a new version or 
2. failed to upgrade to a new version (due to a crash triggering a rollback, or an install failure). 

Therefore, I think we should report DeploymentSucceeded during `notifyApplicationReady()`, if we see that the currentPackage has isFirstRun set to true, because that is the only time we can ascertain that an app has been installed successfully.

Also, we will need a new mechanism to find out that an app has been rolled back, and call reportStatus() there.

src/android/CodePush.java
```diff
@@ -124,32 +137,36 @@ private boolean execInstall(CordovaArgs args, CallbackContext callbackContext) {
...
+    private boolean execRestartApplication(CordovaArgs args, CallbackContext callbackContext) {
+        try {
+            /* check if we have a deployed package already */
+            CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
+            if (deployedPackageMetadata != null) {
+                callbackContext.success();
+                didStartApp = false;
+                onStart();
+            } else {
+                callbackContext.error("The application has no installed updates.");
```
geof90: It seems like this will not actually restart the app if there is no CodePush installed package. Perhaps this should be documented as it might cause confusion, and also this behaviour should be described in the comments here:
https://github.com/Microsoft/cordova-plugin-code-push/pull/53/files#diff-cdca2bab8aff8d52c97a48f83b4229d0R51

src/android/CodePushPackageManager.java
```diff
@@ -95,4 +95,16 @@ public InstallOptions getPendingInstall() {
     public void clearPendingInstall() {
         this.codePushPreferences.clearPendingInstall();
     }
+
+    public void markUnconfirmedInstall(boolean unconfirmed) {
```
geof90: suggestion: maybe it would be easier to understand if there were two methods, markUnconfirmedInstall, and markConfirmedInstall / clearUnconfirmedInstall, instead of passing the boolean flag to determine what this method does.

www/sdk.ts
```diff
@@ -49,14 +64,14 @@ class Sdk {
     /**
      * Reports the update status to the CodePush server.
      */
-    public static reportStatus(status: string, callback?: Callback<void>) {
+    public static reportStatus(status: string, pkg: IPackage, callback?: Callback<void>) {
```
geof90: The status and pkg arguments should be optional. https://github.com/Microsoft/code-push/pull/101/files#diff-a937c637a47cbd31cbb52c89bef7d197R143 If unspecified, this will report to the server that the app is running a new binary version with no CodePush installed package. See my other comment for more information.

src/android/CodePushPackageManager.java
dlebu: The implementation was done that way because we only care about two states (is confirmed / is not confirmed), and from the `CodePushPackageManager` perspective we only see two states. That is the reason for which I consider a single method with a boolean parameter would do it. Internally we have 3 states (true / false / not set), with the last two states being the same. As long as we are state-consistent in the internal implementation, we should be fine, and we do have consistency.

src/android/CodePush.java
dlebu: I changed the implementation to always restart. I will push the changes in a bit.

www/sdk.ts
dlebu: I changed them to be optional.

www/localPackage.ts
dlebu: @lostintangent @geof90 @dtivel I am working on making this change but I think we need to distinguish between a failed install and a rollback. In case of an error, the developer would like to have more details about what happened.

bin/www/codePush.js
```diff
@@ -26,8 +26,28 @@ var CodePush = (function () {
...
+    };
     CodePush.prototype.getCurrentPackage = function (packageSuccess, packageError) {
-        return LocalPackage.getPackageInfoOrNull(LocalPackage.PackageInfoFile, packageSuccess, packageError);
+        NativeAppInfo.isPendingUpdate(function (pendingUpdate) {
+            if (pendingUpdate) {
+                LocalPackage.getPackageInfoOrNull(LocalPackage.OldPackageInfoFile, packageSuccess, packageError);
+            }
+            else {
+                LocalPackage.getPackageInfoOrNull(LocalPackage.PackageInfoFile, packageSuccess, packageError);
+            }
```
shishirx34: you can refactor the lines 34-39:

var packageInfoFile = pendingUpdate ? LocalPackage.OldPackageInfoFile : LocalPackage.PackageInfoFile
LocalPackage.getPackageInfoOrNull(packageInfoFile, pacakgeSuccess, packageError);

bin/www/localPackage.js
```diff
@@ -91,11 +91,16 @@ var LocalPackage = (function (_super) {
                     else {
                         var invokeSuccessAndInstall = function () {
                             CodePushUtil.logMessage("Install succeeded.");
-                            installSuccess && installSuccess();
-                            cordova.exec(function () { }, function () { }, "CodePush", "install", [deployDir.fullPath, installOptions.rollbackTimeout.toString(), installOptions.installMode.toString()]);
+                            if (installOptions.installMode === InstallMode.IMMEDIATE) {
+                                installSuccess && installSuccess();
+                                cordova.exec(function () { }, function () { }, "CodePush", "install", [deployDir.fullPath, installOptions.installMode.toString()]);
```
shishirx34: does exec always require an function parameter? can it not be null or undefined?

src/android/CodePushPackageManager.java
shishirx34: I agree with @geof90 the name is confusing a bit, if you want to confirm install, we pass markUnconfirmedInstall(false) - double negation.. atleast for me makes it hard to read.

Having two methods would be a lot clearer or may be changing this to ```markInstall(boolean isConfirmed)``` that way when I say markInstall(true) this is a confirmed installl markInstall(false) is a an unconfirmed. If that makes sense. Just a suggestion.

src/android/CodePushPreferences.java
```diff
@@ -67,15 +67,31 @@ public InstallOptions getPendingInstall() {
...
+    public void clearNotConfirmedInstall() {
+        this.clearPreferences(CodePushPreferences.INSTALL_NOT_CONFIRMED_PREFERENCE);
+    }
+
+    public boolean isNotConfirmedInstall() {
+        SharedPreferences preferences = context.getSharedPreferences(CodePushPreferences.INSTALL_NOT_CONFIRMED_PREFERENCE, Context.MODE_PRIVATE);
+        boolean notConfirmedInstall = preferences.getBoolean(CodePushPreferences.INSTALL_NOT_CONFIRMED_KEY, false);
+        return notConfirmedInstall;
+    }
+
```
shishirx34: I don't know if its too much trouble, but can we instead of saving notConfirmed save values like isConfirmed and flip the values. I think it would be easy to follow that way. If its a lot of change and could potentially delay our deliveries, I would leave it.

test/test.ts
```diff
@@ -85,7 +85,15 @@ function setupScenario(scenarioPath: string): Q.Promise<void> {
     app.post("/reportTestMessage", function(req: any, res: any) {
         console.log("Application reported a test message.");
         console.log("Body: " + JSON.stringify(req.body));
-        res.sendStatus(200);
+
+        if (!testMessageResponse) {
+            console.log("Seding OK");
```
shishirx34: Sending

www/codePush.ts
```diff
@@ -34,26 +34,54 @@ class CodePush implements CodePushCordovaPlugin {
...
+     * @param packageError Optional callback invoked in case of an error.
+     */
     public getCurrentPackage(packageSuccess: SuccessCallback<LocalPackage>, packageError?: ErrorCallback): void {
-        return LocalPackage.getPackageInfoOrNull(LocalPackage.PackageInfoFile, packageSuccess, packageError);
+        NativeAppInfo.isPendingUpdate((pendingUpdate: boolean) => {
+            if (pendingUpdate) {
+                LocalPackage.getPackageInfoOrNull(LocalPackage.OldPackageInfoFile, packageSuccess, packageError);
+            } else {
+                LocalPackage.getPackageInfoOrNull(LocalPackage.PackageInfoFile, packageSuccess, packageError);
+            }
```
shishirx34: Oh I think I left a comment to refactor on js, applied here.

src/android/CodePushPackageManager.java
dlebu: I changed the implementation to have two methods: markUnconfirmedInstall() and clearUnconfirmedInstall(), both not taking parameters.

bin/www/codePush.js
dlebu: Done. Thanks!

bin/www/localPackage.js
dlebu: It needs to be defined.

src/android/CodePushPreferences.java
dlebu: The state which is relevant (i.e. needs handling) is the not confirmed one, that's why I used it for the name of the methods. A slightly better name would have been `installNeedsConfirmation()`.

test/test.ts
dlebu: Fixed.

www/codePush.ts
dlebu: No problem. Fixed.

www/localPackage.ts
dlebu: I changed part of the reporting mechanism to be initiated on the native side. This is how it currently works:
- For install failures (unzip failed, start page not found, etc) we initiate reporting in the JS layer.
- For rollbacks, install confirmations and store updates, we initiate it from the native layer. Moreover, when any of these events happen, I cache the event on the native side, and only send it over to the JS layer when the application is paused (loses focus). This way, the user experience is not degraded by telemetry.

One thing I wanted to talk about is the store update scenario. In the current implementation, I send a telemetry report for all store updates but not the initial install. Do we need an event for the first (clean) install as well?

shishirx34:
LGTM

www/codePush.ts
```diff
@@ -34,26 +34,86 @@ class CodePush implements CodePushCordovaPlugin {
...
+    
+    /**
+     * Reports an application status back to the server.
+     * !!! This function is called from the native side, please make changes accordingly. !!!
+     */
+    public reportStatus(status: number, label: string, appVersion: string) {
+        try {
+            console.log("Reporting status: " + status + " " + label + " " + appVersion);
+
+            var createPackageForReporting = (label: string, appVersion: string): IPackage => {
```
geof90: What if a package is installed under a different deployment key that was dynamically supplied during checkForUpdate?

www/codePush.ts
```diff
@@ -34,26 +34,86 @@ class CodePush implements CodePushCordovaPlugin {
...
+     */
+    public restartApplication(installSuccess: SuccessCallback<void>, errorCallback?: ErrorCallback): void {
+        cordova.exec(installSuccess, errorCallback, "CodePush", "restartApplication", []);
+    }
+    
+    /**
+     * Reports an application status back to the server.
+     * !!! This function is called from the native side, please make changes accordingly. !!!
+     */
+    public reportStatus(status: number, label: string, appVersion: string) {
```
geof90: Does this handle the case where a package is installed under a different deployment key that was dynamically supplied during checkForUpdate?

dlebu: I just pushed a change that handles this. It also handles the download reporting. Please take a look.

dlebu:
I pushed a change that adds download and first run reporting.

zwacky:
really need these changes. any ETA on this?

lostintangent:
LGTM!

src/android/CodePushPreferences.java
dlebu: @shishirx34 @geof90 I ended up renaming all the methods to `installNeedsConfirmation()`, `markInstallNeedsConfirmation()`, `clearInstallNeedsConfirmation()` today, with the latest changes.

geof90:
LGTM!

lostintangent:
@zwacky We should be publishing this release today. Sorry for the delay!

PR #54, opened 2016-01-18
Don't ignore installOptions upon install

zwacky:
the parameter is currently always ignored and the default installOptions will be taken.
```
local.install(installSuccess, installError, {installMode: 1});
```

because of this:
```
var installOptions = {installMode: 1};
if (!installOptions) {
    // this will never be called
    console.log('never been called');
}
```

### DIFF
--- a/bin/www/localPackage.js
+++ b/bin/www/localPackage.js
@@ -29,7 +29,7 @@ var LocalPackage = (function (_super) {
         var _this = this;
         try {
             CodePushUtil.logMessage("Installing update package ...");
-            if (!installOptions) {
+            if (!installOptions || Object.keys(installOptions).length == 0) {
                 installOptions = LocalPackage.getDefaultInstallOptions();
             }
             else {
--- a/www/localPackage.ts
+++ b/www/localPackage.ts
@@ -11,8 +11,8 @@ import CodePushUtil = require("./codePushUtil");
 import Sdk = require("./sdk");
 
 /**
- * Defines a local package. 
- * 
+ * Defines a local package.
+ *
  * !! THIS TYPE IS READ FROM NATIVE CODE AS WELL. ANY CHANGES TO THIS INTERFACE NEEDS TO BE UPDATED IN NATIVE CODE !!
  */
 class LocalPackage extends Package implements ILocalPackage {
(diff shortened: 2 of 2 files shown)


dlebu:
Hi @zwacky, thanks for reporting the issue. 
From your description, my understanding of the problem is that the `installMode` parameter is always ignored during installation. I tried to reproduce the problem but it seems to work fine in production code. 

[The line you modified](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/localPackage.ts#L57) handled that specific scenario. If the `installOptions` object evaluates to `false`, we do take the default options, which is expected. This is the case where the parameter was not specified, or it was `null`, `undefined`. In this scenario, we would like the default options.

If it does not evaluate to false (say it is `{installMode: 1}`, similar to your example), [we copy](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/localPackage.ts#L60) the unassigned parameters of the `installOptions` object from the default options. For the case of your example, we don't copy anything since the only parameter of `installOptions` which we need is `installMode`, which is already assigned to `1`. So we would end up using `{installMode: 1}`, which is what the developer passed as a parameter.

[Your fix](https://github.com/Microsoft/cordova-plugin-code-push/pull/54/files#diff-55b3fc1863700c061624c1967cf97d83R57) seems to handle the case where the user passed in an empty object (`{}`), but that is already handled by the second case. For that scenario, we will indeed copy `installMode` from the default options. We would end up using `{installMode: 1}`, which is the default install mode.

Please let me know if I am missing something and if you still think the parameter handling needs the fix.

zwacky:
my example might have been unclear.

I want to install the update with the flag `InstallMode.ON_NEXT_RESUME`.

```
localPackage.install(onInstallSuccess, onInstallError, {
    installMode: window.InstallMode.ON_NEXT_RESUME
});
```

but the installMode parameter gets ignored, because `({installMode: 2}) === false` described [here](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/localPackage.ts#L57).

dlebu:
`!!{installMode: 2}` evaluates to `true`, which would make `!{installMode: 2}` evaluate to `false`, right? In the case of your example, the options parameter would remain `{installMode: 2}`. If you think there is a problem with the installation on resume, please provide us with more feedback - the problem would most likely be somewhere else in the code.

For a quick check, try to evaluate the your last expression `({installMode: 2}) === false` in a browser. It will evaluate to `false`.

zwacky:
arg. you're totally right. dunno what approach i was going for... :S

PR #55, opened 2016-01-18
Add missing tslint because gulp-tslint doesn't include it

zwacky:


dlebu:
@zwacky Thanks for the contribution!

PR #66, opened 2016-02-29
Readme updates

lostintangent:
This PR updates the README primarily in three ways:

1. Updates the getting started section to call out the need for the `cordova-plugin-whitelist` plugin, since we've seen issues with apps that didn't already have it installed.

1. Adds a note about App Store policies, so that devs can be more informed about their usage of the CodePush SDK (we already added this note to the React Native readme)

1. Adds a note about how to upgrade a Cordova platform in the event that a user is running an unsupported version

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,21 @@ Cordova 5.0.0+ is fully supported, along with the following asociated platforms:
 - Android ([cordova-android](https://github.com/apache/cordova-android) 4.0.0+)
 - iOS ([cordova-ios](https://github.com/apache/cordova-ios) 3.9.0+)
 
-To check which versions of each Cordova platform you are currently using, you can run the following command:
+To check which versions of each Cordova platform you are currently using, you can run the following command and inspect the `Installed platforms` list:
 
 ```shell
-cordova platform version
+cordova platform
+```
+
+If you're running an older Android and/or iOS platform than is mentioned above, and would be open to upgrading, you can easily do so by running the following commands (omitting a platform if it isn't neccessary):
+
+```shell
+cordova platform upgrade android
+cordova platform upgrade ios
 ```
 
 ## Getting Started
+
 Once you've followed the general-purpose ["getting started"](http://microsoft.github.io/code-push//docs/getting-started.html) instructions for setting up your CodePush account, you can start CodePush-ifying your Cordova app by running the following command from within your app's root directory:
 
 ```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,9 +65,22 @@ With the CodePush plugin installed, configure your app to use it via the followi
     <meta http-equiv="Content-Security-Policy" content="default-src https://codepush.azurewebsites.net 'self' data: gap: https://ssl.gstatic.com 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src *" />
     ```
    
+4. Finally, double-check that you already have the [`cordova-plugin-whitelist`](https://github.com/apache/cordova-plugin-whitelist) plugin installed (most apps will). To check this, simply run the following command:
+
+    ```shell
+    cordova plugin
```
dlebu: `cordova plugin list` is the actual command, although the `cordova plugin` command defaults to that as well.

README.md
```diff
@@ -57,9 +65,22 @@ With the CodePush plugin installed, configure your app to use it via the followi
...
+    ```
+    
+    If `cordova-plugin-whitelist` is in the list, then you are good to go. Otherwise, simply run the following command to add it:
+    
+    ```shell
+    cordova plugin cordova-plugin-whitelist
```
dlebu: `cordova plugin add cordova-plugin-whitelist`

README.md
```diff
@@ -74,6 +95,8 @@ codePush.sync();
 
 If an update is available, it will be silently downloaded, and installed the next time the app is restarted (either explicitly by the end user or by the OS), which ensures the least invasive experience for your end users. If you would like to display a confirmation dialog (an "active install"), or customize the update experience in any way, refer to the sync method's API reference for information on how to tweak this default behavior.
 
+*NOTE: While [Apple's developer agreement](https://developer.apple.com/programs/ios/information/iOS_Program_Information_4_3_15.pdf) fully allows performing over-the-air updates of JavaScript and assets (which is what enables CodePush!), it is against their policy for an app to display an update prompt. Because of this, we recommend that App Store-distributed apps don't enable the `updateDialog` option when calling `sync`, whereas Google Play and internally distributed apps (e.g. Enterprise, Fabric, HockeyApp) can choose to enable/customize it.*
```
dlebu: Is this the case just for the first version or all versions? Can we advise they disable the update dialog only on review submission?

README.md
```diff
@@ -14,13 +14,21 @@ Cordova 5.0.0+ is fully supported, along with the following asociated platforms:
...
 
 ```shell
-cordova platform version
+cordova platform
+```
+
+If you're running an older Android and/or iOS platform than is mentioned above, and would be open to upgrading, you can easily do so by running the following commands (omitting a platform if it isn't neccessary):
+
+```shell
+cordova platform upgrade android
```
dlebu: What is the minimum version of the CLI supporting this? I just tried it on 5.4.0 and it doesn't seem to work.

README.md
lostintangent: Devs can definitely release an update that enables the dialog and/or always use the dialog and just make sure to not release to CodePush until after the app is approved, but I'm not sure if I want to make that recommendation here :)

We know for a fact that the update dialog is against Apple's policies, and we've now seen 5 apps rejected because of it (they were all approved after removing the dialog), so I want to be really conservative in the prescription we give.

Since you already "convert" mandatory updates to immediately install, the Cordova plugin has a pretty solid "silent mode" experience, and I want to mirror that behavior in the RN plugin to ensure that we optimize for apps that don't use the dialog, but want to be able to maintain the semantics of a mandatory update.

README.md
lostintangent: Good question, I thought this was the standard behavior. Should we simply document doing an rm and then an add instead?

dlebu: I just checked the doc. The command is `update`, not upgrade. :+1:

lostintangent: Doh! That is what I meant to add :)

dlebu:
LGTM! Thanks for making this change!

Issue #60, opened 2016-01-31
Cordova WKWebview Support


tropperstyle:
I am pulling out my hair trying to find a Cordova solution that will allow for hot code updates, as well as take advantage of the vastly more performant WKWebview. I have tried the following plugins without much luck. I realize since each of these plugins alter the way the application code is loaded, compatibility may not be so trivial. Has there been any experimental setups? Any insight or at least nudges in the right direction would be much appreciated!

https://github.com/Telerik-Verified-Plugins/WKWebView
https://github.com/apache/cordova-plugin-wkwebview-engine
https://github.com/apache/cordova-plugins/tree/master/wkwebview-engine-localhost

Not sure if this helps, but here is a related issue with a similar plugin to code push https://github.com/nordnet/cordova-hot-code-push/issues/42

lostintangent:
Hey @tropperstyle! Unfortunately, we haven't tested CodePush with either of these plugins, so I couldn't offer any immediate insights, but I can definitely take a look at it on Monday and see what kind of issues exist and what workarounds we could prescribe. We definitely need this scenario to work.

Stay tuned and sorry for the inconvenience!

tropperstyle:
Any update on this? Looks like `cordova-plugin-wkwebview-engine` is a core plugin now so hopefully that makes hooking in a bit easier

lostintangent:
I ended up having to look into a few other issues today unfortunately. I'll definitely get to this in the next day or two.

tropperstyle:
Thanks for the update. I appreciate the time and effort!

tropperstyle:
Sorry to nag, but any movement on this? I can attempt to hack it together but I don't have much experience on the native side.

lostintangent:
I started looking at it today and will hopefully finish up the analysis tomorrow of what we'll need to change in our plugin to support it. It doesn't look too difficult so far.

dozoisch:
Any update on this :)? I was thinking about migrating to wkwebview too!

lostintangent:
@dozoisch We've got this work scheduled for the upcoming week. We did the analysis to figure out what was necessary to support this, so we'll try to get a plugin update published within the week and we'll ping you when it's out.

tropperstyle:
[...]

PR #69, opened 2016-03-01
Fix wkwebview compat

ryuyu:
Update the plugin to be compatible with both UIWebView and WKWebView.

Addresses https://github.com/Microsoft/cordova-plugin-code-push/issues/60

@dlebu @lostintangent

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -124,7 +124,9 @@ gulp.task("default", function (callback) {
     runSequence("clean", "compile", "tslint", callback);
 });
 
-gulp.task("test-ios",["test-ios-uiwebkit", "test-ios-wkwebkit"]);
+gulp.task("test-ios", function (callback) {
+    runSequence("test-ios-uiwebview", "test-ios-wkwebview", callback);
+});
 
 gulp.task("test-ios-uiwebview", function (callback) {
     var command = "mocha";
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -124,9 +124,17 @@ gulp.task("default", function (callback) {
     runSequence("clean", "compile", "tslint", callback);
 });
 
-gulp.task("test-ios", function (callback) {
+gulp.task("test-ios",["test-ios-uiwebkit", "test-ios-wkwebkit"]);
+
+gulp.task("test-ios-uiwebview", function (callback) {
     var command = "mocha";
-    var args = ["./bin/test", "--mockserver", "http://127.0.0.1:3000", "--platform", "ios"];
+    var args = ["./bin/test", "--mockserver", "http://127.0.0.1:3000", "--platform", "ios", "--use-wkwebview", "false"];
+    executeCommand(command, args, callback);
+});
+
+gulp.task("test-ios-wkwebview", function (callback) {
+    var command = "mocha";
+    var args = ["./bin/test", "--mockserver", "http://127.0.0.1:3000", "--platform", "ios", "--use-wkwebview", "true"];
     executeCommand(command, args, callback);
 });
 
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -221,7 +221,7 @@ bool pendingInstall = false;
 }
 
 - (void)applicationWillResignActive {
-    [Reporting reportStatuses:((UIWebView*)((CDVViewController *)self.viewController).webView)];
+    [Reporting reportStatuses:self.webView];
 }
 
 - (BOOL)loadPackage:(NSString*)packageLocation {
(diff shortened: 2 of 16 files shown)


src/ios/Reporting.m
```diff
@@ -1,3 +1,4 @@
+#import <Webkit/Webkit.h>
```
lostintangent: Would this work correctly on older iOS versions? I think the `WKWebView` is only available on iOS 8+, and I'm not sure if we're ready to drop iOS 7 support completely.

src/ios/Reporting.h
```diff
@@ -1,3 +1,5 @@
+#import <Webkit/Webkit.h>
```
shishirx34: do you need this import?

src/ios/Reporting.m
ryuyu: We can get rid of this import by dynamically calling the WKWebView Javascript executor if the object responds to the selector.

src/ios/Reporting.h
ryuyu: No. I removed it and updated the code to dynamically call the methods on it.

tropperstyle:
Thanks for this!! But I ran into issues with `codePush.sync()`

XMLHttpRequest cannot load https://codepush.azurewebsites.net/updateCheck?deploymentKey=MYKEY&appVersion=1.0.0&packageHash=&isCompanion=false&label=. Origin null is not allowed by Access-Control-Allow-Origin.

I have added `https://codepush.azurewebsites.net` to my `<meta http-equiv="Content-Security-Policy"/>` tag as well as `<access origin="https://codepush.azurewebsites.net" />` and `<access origin="https://codepush.blob.core.windows.net" />` in config.xml.

I don't get the error after removing the `cordova-plugin-wkwebview-engine` plugin. It could be similar to [this issue](https://github.com/Telerik-Verified-Plugins/WKWebView/issues/87#issue-62983288).

lostintangent:
@trooperstyle There is a server change that needs to go along with this that supports an odd CORS behavior that WKWebView has. We're going to deploy that change tomorrow morning, at which point, you could grab these bits and go :)

tropperstyle:
Ah, too trigger happy. Thanks again!

src/ios/CodePush.m
```diff
@@ -235,7 +235,9 @@ - (BOOL)loadPackage:(NSString*)packageLocation {
 }
 
 - (void)loadURL:(NSURL*)url {
-    [((UIWebView*)((CDVViewController *)self.viewController).webView) loadRequest:[NSURLRequest requestWithURL:url]];
+    // Cast to a UIWebView here to enable call to loadRequest.
+    // Both UIWebView and WKWebview share this call, so regardless of which type the returned webview is, this will work.
+    [(UIWebView*)self.webView loadRequest:[NSURLRequest requestWithURL:url]];
```
dlebu: What versions of cordova-ios did you use for testing? We currently support 3.9.0+ so we need to make sure this will compile on all versions. I remember we had a compilation issue with 4.0.1 in the past.

dlebu:
Thanks for making this fix. It looks good, I had one comment about `cordova-ios` versions. Can you add at least one test that uses `WKWebView`?

src/ios/CodePush.m
ryuyu: Ah. I don't think i tested with 3.9.0. I'll hit that today and see if this works.

ryuyu: This compiles correctly on 3.9.0+. The older version that referenced webkit did not compile, but we removed that direct reference.

lostintangent:
@tropperstyle The server bits have been deployed, so if you'd like to pull the bits from GitHub you should be unblocked. Additionally, we'll try to get this update deployed to NPM within the next day or two.

lostintangent:
LGTM!

gulpfile.js
```diff
@@ -124,9 +124,17 @@ gulp.task("default", function (callback) {
     runSequence("clean", "compile", "tslint", callback);
 });
 
-gulp.task("test-ios", function (callback) {
+gulp.task("test-ios",["test-ios-uiwebkit", "test-ios-wkwebkit"]);
```
dlebu: The task name is wrong here: `test-ios-uiwebview`. Also, if I remember correctly, due to paths in tests we cannot run them in parallel so I would make them run in sequence like this:
```javascript
gulp.task("test-ios", function (callback) {
    runSequence("test-ios-uiwebview", "test-ios-wkwebview", callback);
});
```

dlebu:
Minor comment on gulp task, otherwise LGTM. Thanks again for making this fix!

Issue #65, opened 2016-02-18
FileUploadOption in fileTransfer.d.ts


podger:
Hi,
I've encountered an issue trying to compile my TypeScript code envolving the cordova file transfer plugin.
```
interface FileUploadOptions {
    ...

    /** A map of header name/header values. Use an array to specify more than one value. */
    headers?: Object[];
}
```

Shouldn't it be:
`    headers?: Object;
`
instead? This way is working fine. I looked up the [File Transfer Docs](https://cordova.apache.org/docs/it/3.1.0/cordova/file/filetransfer/filetransfer.html), (even if it's not updated) and doesn't mention array object in the headers param.

I guess the same is for `FileDownloadOptions` 

Am i missing something?

Thank you,
Bye

lostintangent:
@podger If you change the d.ts to use an object instead of an object array, does your app compile then?

podger:
@lostintangent yes it compiles perfectly using an `Object `type using tsc 1.0.3.0, gives error instead with `Object[]`

lostintangent:
Resolved by #74

PR #74, opened 2016-03-10
Update fileTransfer.d.ts

ryuyu:
Fixing incorrect typings in the fileTransfer.d.ts (https://github.com/Microsoft/cordova-plugin-code-push/issues/65)

Both of these interfaces are defined as taking headers as an object array, but this is incorrect.
Quickly checking https://github.com/apache/cordova-plugin-file-transfer/blob/master/src/android/FileTransfer.java shows that they both take headers as a single object and not an array.

@dlebu @lostintangent

### DIFF
--- a/typings/fileTransfer.d.ts
+++ b/typings/fileTransfer.d.ts
@@ -93,13 +93,13 @@ interface FileUploadOptions {
     /** Whether to upload the data in chunked streaming mode. Defaults to true. */
     chunkedMode?: boolean;
     /** A map of header name/header values. Use an array to specify more than one value. */
-    headers?: Object[];
+    headers?: Object
 }
 
 /** Optional parameters for download method. */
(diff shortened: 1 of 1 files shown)


lostintangent:
LGTM!

typings/fileTransfer.d.ts
```diff
@@ -93,13 +93,13 @@ interface FileUploadOptions {
     /** Whether to upload the data in chunked streaming mode. Defaults to true. */
     chunkedMode?: boolean;
     /** A map of header name/header values. Use an array to specify more than one value. */
-    headers?: Object[];
+    headers?: Object
```
shishirx34: semi colon

dlebu:
:+1:

shishirx34:
Looks good.

PR #77, opened 2016-03-15
Changing default dialog strings

lostintangent:
This simply changes two of the default dialog strings in order to match the existing behavior with the React Native plugin.


dtivel:
Looks good.

PR #78, opened 2016-03-15
Adding more logging

lostintangent:
This PR updates some existing log messages to remove the use of the word "package" and be more consistent with the RN plugin logs. Additionally, I introduced new logs for the following scenarios:

1. Checking for an update, this way, a dev can see this is occurring, as opposed to needing to wait for the check to complete and display the up-to-date or update-available message

2. After an update is installed, we display a log indicating when it will be applied.

3. If an available update is being ignored due to having been blacklisted.

4. When displaying an update prompt, and if the user cancelled/ignored it.

5. If an available update is being ignored due to it targeting a newer binary version.

Additionally, this includes the JS files for #77, which I forgot to check-in.

### DIFF
--- a/bin/www/codePush.js
+++ b/bin/www/codePush.js
@@ -75,11 +75,12 @@ var CodePush = (function () {
                 }
                 else {
                     var appUpToDate = function () {
-                        CodePushUtil.logMessage("The application is up to date.");
+                        CodePushUtil.logMessage("App is up to date.");
                         querySuccess && querySuccess(null);
                     };
                     if (remotePackageOrUpdateNotification) {
                         if (remotePackageOrUpdateNotification.updateAppVersion) {
+                            CodePushUtil.logMessage("An update is available, but it is targetting a newer binary version than you are currently running.");
                             appUpToDate();
                         }
                         else {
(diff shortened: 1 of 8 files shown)


bin/www/codePush.js
```diff
@@ -75,11 +75,12 @@ var CodePush = (function () {
                 }
                 else {
                     var appUpToDate = function () {
-                        CodePushUtil.logMessage("The application is up to date.");
+                        CodePushUtil.logMessage("App is up to date.");
                         querySuccess && querySuccess(null);
                     };
                     if (remotePackageOrUpdateNotification) {
                         if (remotePackageOrUpdateNotification.updateAppVersion) {
+                            CodePushUtil.logMessage("An update is available, but it is targetting a newer binary version than you are currently running.");
```
shishirx34: targeting

www/codePush.ts
```diff
@@ -133,13 +133,14 @@ class CodePush implements CodePushCordovaPlugin {
...
                     var appUpToDate = () => {
-                        CodePushUtil.logMessage("The application is up to date.");
+                        CodePushUtil.logMessage("App is up to date.");
                         querySuccess && querySuccess(null);
                     };
 
                     if (remotePackageOrUpdateNotification) {
                         if ((<NativeUpdateNotification>remotePackageOrUpdateNotification).updateAppVersion) {
                             /* There is an update available for a different version. In the current version of the plugin, we treat that as no update. */
+                            CodePushUtil.logMessage("An update is available, but it is targetting a newer binary version than you are currently running.");
```
shishirx34: targeting

dlebu:
LGTM! Please run the tests before you check in. :+1:

shishirx34:
LGTM

PR #81, opened 2016-03-26
Prevent sync from being called multiple times concurrently

scottbommarito:
Added a static variable to remember whether or not a sync is ongoing, and if another sync starts before that sync finishes, make it return immediately with a new SyncStatus, "IN_PROGRESS". Also added test cases to test each sync functionality with sync called twice.


www/codePush.ts
```diff
@@ -187,6 +191,7 @@ class CodePush implements CodePushCordovaPlugin {
     /**
      * Convenience method for installing updates in one method call.
      * This method is provided for simplicity, and its behavior can be replicated by using window.codePush.checkForUpdate(), RemotePackage's download() and LocalPackage's install() methods.
+     * If another sync is already running, it yields SyncStatus.SYNC_IN_PROGRESS.
```
lostintangent: SyncStatus.IN_PROGRESS

test/test.ts
```diff
@@ -799,130 +801,255 @@ describe("window.codePush", function() {
...
+                        noUpdateReponse.isAvailable = false;
+                        noUpdateReponse.appVersion = "0.0.1";
+                        mockResponse = { updateInfo: noUpdateReponse };
+                        testMessageCallback = verifyMessages([new su.AppMessage(su.TestMessage.SYNC_STATUS, [su.TestMessage.SYNC_IN_PROGRESS])],
+                            deferred);
+                        projectManager.restartApplication(targetPlatform, TestNamespace, testRunDirectory, targetEmulator).done();
+                        return deferred.promise;
+                    })
+                    .done(done, done);
+            });
```
geof90: I'm a bit confused as to what the goal of these tests are. It seems all of them are really only testing the same thing (that the second sync returns SYNC_IN_PROGRESS). Even though different scenarios are being run, we are not verifying the output of the first sync to verify that they are succeeding.

I would recommend either just having one test to test `IN_PROGRESS` specifically, or modifying the existing test to check for the expected output of the first `sync()` call as well.

scottbommarito: The goal is to verify that the second sync doesn't affect the outcome of the first sync, but I see what you mean. I'll add the result state of the first sync to all these tests just so we know the first sync finished as expected.

test/template/www/js/updateSync2x.js
```diff
@@ -0,0 +1,35 @@
...
+    },
+    onDeviceReady: function () {
+        app.receivedDeviceReady();
+    },
+    // Update DOM on a Received Event
+    receivedDeviceReady: function () {
+        document.getElementById("deviceready").innerText = "Device is ready (scenario - update sync)";
+        console.log('Received Event: deviceready');
+        app.sendTestMessage("DEVICE_READY_AFTER_UPDATE");
+        /* invoke sync with UI options such that the update will not be installed */
```
geof90: Does this comment make sense? We aren't invoking it with any UI options.

scottbommarito: not sure, it was in the original updateSync.js and I didn't change it, we would need to change it there too

scottbommarito: i think it might be referring to how it's using the default install mode instead of IMMEDIATE, which all the other tests use

test/test.ts
scottbommarito: This has been updated now--all the tests include the whole sync status output from both syncs.

geof90:
LGTM! @lostintangent @dlebu any issues with this getting merged?

www/codePush.ts
```diff
@@ -206,7 +211,58 @@ class CodePush implements CodePushCordovaPlugin {
...
+            var syncCallbackAndUpdateSyncInProgress: SuccessCallback<any> = (result: any): void => {
+                switch (result) {
+                    case SyncStatus.ERROR:
+                    case SyncStatus.IN_PROGRESS:
+                    case SyncStatus.UP_TO_DATE:
+                    case SyncStatus.UPDATE_IGNORED:
+                    case SyncStatus.UPDATE_INSTALLED:
+                        /* The sync has completed */
+                        CodePush.SyncInProgress = false;
+                        break;
```
lostintangent: Do we even need this break or the following case statements? This should work as expected without them.

scottbommarito: i was under the impression that every case needs to exit with a break at some point, so i could take out the intermediate states and only have a break at the end

lostintangent: I would vote to remove it simply so that we don't have code to maintain (e.g. when we add a new `SyncStatus`), that doesn't actually add any value.

lostintangent:
LGTM!

PR #83, opened 2016-03-28
Add .gitignore

scottbommarito:
Previously, there was no .gitignore file in the repo, so whenever you would build, about 6 thousand changed files would appear in GitHub for Windows. This is incredibly cluttery and slow, so I added a .gitignore that ignores files in node_modules, which are built by "npm install", and files in bin/test, which are built by "gulp" for testing purposes


geof90:
LGTM!

PR #88, opened 2016-03-30
Hash binary contents

geof90:
This PR adds support for

- Generating a hash of the binary's www contents
- Sending that hash during updateCheck requests, to prevent having to acquire and install an update that is exactly the same as the version shipped in the binary
- Applying a diff update against the binary's contents if necessary


src/android/CodePush.java
```diff
@@ -242,6 +273,7 @@ private void returnStringPreference(String preferenceName, CallbackContext callb
     }
 
     private void handleAppStart() {
+        this.isRunningBinaryVersion = true;
```
shishirx34: is `handleAppStart()` called multiple times? If once, do still need this given that you instantiated it to `true` above?

src/ios/CodePush.m
```diff
@@ -7,11 +7,40 @@
...
+        pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK
+                                         messageAsString:binaryHash];
+    } else {
+        NSError* error;
+        binaryHash = [UpdateHashUtils getBinaryHash:&error];
+        if (error) {
+            pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR
+                                             messageAsString:[@"An error occurred when trying to get the hash of the binary contents. " stringByAppendingString:error.description]];
+        } else {
+            pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK
```
shishirx34: shouldn't you call `saveBinaryHash(binaryHash)` before here?

src/ios/CodePush.m
```diff
@@ -164,6 +193,7 @@ - (void)dealloc {
 }
 
 - (void)handleAppStart {
+    isRunningBinaryVersion = true;
```
shishirx34: same as above, is this needed?

src/ios/CodePushPackageManager.h
```diff
@@ -8,6 +8,7 @@
 + (CodePushPackageMetadata*)getCurrentPackageMetadata;
 + (void)clearFailedUpdates;
 + (void)cleanDeployments;
++ (NSString*)getCachedBinaryHash;
```
shishirx34: `...(void)saveBinaryHash;` ?

www/localPackage.ts
```diff
@@ -382,15 +396,27 @@ class LocalPackage extends Package implements ILocalPackage {
...
+                NativeAppInfo.getBinaryHash((binaryHashError: Error, binaryHash: string) => {
+                    /** 
+                     * For the default package we need the app version, 
+                     * and ideally the hash of the binary contents.
+                     */
+                    if (appVersionError) {
+                        CodePushUtil.logError("Could not get application version." + appVersionError);
+                        packageError(appVersionError);
+                        return;
+                    }
```
shishirx34: may be move `if (appVersionError) {...}` outside(before) of this `NativeAppInfo.getBinaryHash(...)` function call?

src/android/CodePush.java
```diff
@@ -42,10 +45,15 @@ public boolean execute(String action, CordovaArgs args, final CallbackContext ca
         } else if ("getDeploymentKey".equals(action)) {
             this.returnStringPreference("codepushdeploymentkey", callbackContext);
             return true;
+        } else if ("isRunningBinaryVersion".equals(action)) {
+            callbackContext.success(isRunningBinaryVersion ? 1 : 0);
```
shishirx34: why return a number and not a `boolean` here?

www/nativeAppInfo.ts
```diff
@@ -105,6 +115,21 @@ class NativeAppInfo {
 
         cordova.exec(win, fail, "CodePush", "isPendingUpdate", []);
     }
+    
+    /**
+     * Checks with the native side if it is running the version of the app contained in the binary.
+     */
+    public static isRunningBinaryVersion(callback: SuccessCallback<boolean>): void {
+        var win = (runningBinaryVersion?: number) => {
+            callback(!!runningBinaryVersion);
```
shishirx34: why not return a boolean value from exec? Any issues from iOS side?

src/ios/CodePush.m
```diff
@@ -7,11 +7,40 @@
...
 
 @implementation CodePush
 
 bool didUpdate = false;
 bool pendingInstall = false;
+bool isRunningBinaryVersion = true;
+
+- (void)isRunningBinaryVersion:(CDVInvokedUrlCommand *)command {
+    CDVPluginResult* pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK
+                                                         messageAsInt:isRunningBinaryVersion ? 1 : 0];
```
shishirx34: can't you return true/false here?

src/ios/CodePushPackageManager.h
geof90: added

geof90:
@shishirx34 I decided to completely remove the need for `isRunningBinaryVersion` because I realized that it is vulnerable to a corner case as follows: 

1. app is on binary version
2. app downloads an update but does not restart immediately
3. while running the same binary version, app downloads another update, but this time it erroneously diffs against the binary version due to `isRunningBinaryVersion` still being true.

I decided instead to rely on `getPackage(LocalPackage.PackageInfoFile...)`, if there is no current package, we diff against the version in the binary. The RN plugin uses the same logic.

src/android/UpdateHashUtils.java
```diff
@@ -0,0 +1,73 @@
...
+        MessageDigest messageDigest = null;
+        DigestInputStream digestInputStream = null;
+        try {
+            messageDigest = MessageDigest.getInstance("SHA-256");
+            digestInputStream = new DigestInputStream(dataStream, messageDigest);
+            byte[] byteBuffer = new byte[1024 * 8];
+            while (digestInputStream.read(byteBuffer) != -1);
+        } catch (NoSuchAlgorithmException e) {
+            // Should not happen.
+            throw new RuntimeException("Unable to compute hash of update contents.", e);
```
dlebu: This will probably kill the application if it happens. We could add `NoSuchAlgorithmException` to the list of thrown exceptions and handle it properly in the caller code. I think for now we only handle `IOException` there.

src/android/build-extras.gradle
```diff
@@ -0,0 +1,50 @@
+// Needed to speed up asset listing: refer to https://github.com/apache/cordova-plugin-file/blob/76c8fd5e43598ea660c2db64eb08e2659b45a575/src/android/AssetFilesystem.java#L41
```
dlebu: Do we have any numbers on how slow the asset listing is with/without this extra build step?

shishirx34:
LGTM!

dlebu:
LGTM!

src/android/build-extras.gradle
geof90: Asset listing did not show a significant difference, but recursive copying of the www contents in a blank Cordova app with CodePush added took 4.26 seconds with the index and 7.83 seconds without the index.

Issue #89, opened 2016-03-31
deploymentKey override not working


dylanvdmerwe:
I have left my config.xml file with blank strings for the CodePushDeploymentKey. In code I then determine what the appropriate deploymentKey is and then use the [syncOptions](http://microsoft.github.io/code-push/docs/cordova.html#codepushsync) method to set the key. 

When I run the app, the correct key is set. I get the "Checking for update" status. Then in the logs I see:
> ERROR: [CodePush]  Could not get the CodePush configuration. Please check your config.xml file. 
> ERROR: [CodePush]  An error occurred during sync. 

I thought the whole point of setting the value in the syncOptions was to manually be able to set the key?

```
 <platform name="android">
        <preference name="CodePushDeploymentKey" value="" />

 <platform name="ios">
        <preference name="CodePushDeploymentKey" value="" />
```

```
    getCodePushKey(): string {
        if (this.device.isAndroid) {
            if (AppConstants.IsDebugMode)
                return ''; // staging
            else
                return ''; // production
        }
        else if (this.device.isIos) {
            if (AppConstants.IsDebugMode)
                return ''; // staging
            else
                return ''; // production
        }

        return undefined;
    }


  let syncOptions = {
                    installMode: InstallMode.IMMEDIATE,
                    updateDialog: false,
                    deploymentKey: this.getCodePushKey(),
                };
                window.codePush.sync(this.syncStatusCallback.bind(this), syncOptions, this.syncStaticDownloadProgress.bind(this));
```

dylanvdmerwe:
Seems to be iOS specific.

geof90:
@dylanvdmerwe https://github.com/Microsoft/cordova-plugin-code-push/pull/90 is out for review that fixes this issue, in the meantime while that gets reviewed+merged+published, the workaround is to specify a fake but non-empty value in the config.xml so that it doesn't short circuit the sync process.

dylanvdmerwe:
Epic thanks @geof90 - will give it a try!

PR #90, opened 2016-03-31
Fix deployment key override

geof90:
Fixes https://github.com/Microsoft/cordova-plugin-code-push/issues/89. If a deployment key is not specified in the config.xml, we will still proceed if one was provided during the call to `checkForUpdate()`/`sync()`.


lostintangent:
LGTM!

PR #94, opened 2016-04-07
Add MinimumBackgroundDuration to InstallMode.ON_NEXT_RESUME

scottbommarito:
followed the implementation of this on the react side fairly closely


src/android/CodePush.java
```diff
@@ -436,7 +442,9 @@ public void onStart() {
             /* The application was resumed from the background. */
             /* Handle ON_NEXT_RESUME pending installations. */
             InstallOptions pendingInstall = this.codePushPackageManager.getPendingInstall();
-            if ((pendingInstall != null) && (InstallMode.ON_NEXT_RESUME.equals(pendingInstall.installMode))) {
+            if (pendingInstall != null && 
+                InstallMode.ON_NEXT_RESUME.equals(pendingInstall.installMode) && 
+                (new Date().getTime() - lastPausedTimeMs) / 1000 >= pendingInstall.minimumBackgroundDuration) {
```
dlebu: I would extract this logic to a utility method.

dlebu:
Please add some tests. Basic scenarios from the top of my head:

- developer did not specify a minimum background duration (default value is used)
- app is in the background for more/less time than the value specified and behaves as expected
- other install modes are not affected by the setting

dlebu:
Also, don't forget to update the readme file with the new option.

src/android/CodePush.java
scottbommarito: I would agree if we used this logic multiple times, but it's only here once and on the iOS native side once. I can calculate the difference outside of the if statement if that makes it cleaner (`int backgroundDuration = (new Date().getTime() - lastPausedTimeMs);`) but I don't think an entire function is necessary.

README.md
```diff
@@ -565,6 +565,8 @@ Otherwise, the install operation will be marked as failed, and the application i
 
     - __installMode__: Used to specify the [InstallMode](#installmode) used for the install operation. Defaults `InstallMode.ON_NEXT_RESTART`.
 
+    - __minimumBackgroundDuration__: If __installMode__ is `InstallMode.ON_NEXT_RESUME`, used to specify the amount of time the app must be in the background before the update is installed when it is resumed. Defaults `0`.
```
lostintangent: Can you add this to the docs for `sync` as well?

test/test.ts
```diff
@@ -1133,5 +1136,109 @@ describe("window.codePush", function() {
...
+                            new su.AppMessage(su.TestMessage.SYNC_STATUS, [su.TestMessage.SYNC_UPDATE_INSTALLED])], deferred);
+                        console.log("Running project...");
+                        projectManager.runPlatform(testRunDirectory, targetPlatform, true, targetEmulator).done();
+                        return deferred.promise;
+                    })
+                    .then<void>(() => {
+                        var noUpdateResponse = createDefaultResponse();
+                        noUpdateResponse.isAvailable = false;
+                        noUpdateResponse.appVersion = "0.0.1";
+                        mockResponse = { updateInfo: noUpdateResponse };
```
dlebu: You missed the `verifyMessages` call in this case. Is it on purpose?

README.md
dlebu: Can you also add a small code example using this option?

test/template/www/js/scenarioSyncRestart30.js
```diff
@@ -0,0 +1,47 @@
...
+                    case SyncStatus.ERROR:
+                    case SyncStatus.IN_PROGRESS:
+                        app.sendTestMessage("SYNC_STATUS", [status]);
+                    default:
+                        break;
+                }
+            },
+            {
+                installMode: InstallMode.ON_NEXT_RESTART,
+                minimumBackgroundDuration: 30
```
dlebu: Totally optional, but I would go for smaller values, to avoid unnecessary increases in test time. It will save you time in the long run. I would use values like (5s, 10s, 15s) instead of (10s, 30s, 40s).

test/projectManager.ts
```diff
@@ -166,6 +166,36 @@ export class ProjectManager {
             return ProjectManager.runPlatform(testRunDirectory, targetPlatform, true, targetEmulator);
         }
     }
+    
+    /**
+     * Navigates away from the application and then navigates back to it.
+     */
+    public static resumeApplication(delayBeforeResumingMs: number, targetPlatform: platform.IPlatform, namespace: string, testRunDirectory: string, targetEmulator: string): Q.Promise<void> {
+        var emulatorManager = targetPlatform.getOptionalEmulatorManager();
```
dlebu: This class is intended to be platform agnostic. Most of its functionality relies on the Cordova CLI. Can you add a `resumeApplication` method to the `IEmulatorManager` interface and implement if for both Android and iOS there, in `platform.ts`? I would create an `EmulatorManager` for Android as well (currently the android platform doesn't have one) and move this code here. The iOS specific code can go in `IOSEmulatorManager ` and the Android specific in a new `AndroidEmulatorManager` class.

src/android/CodePush.java
dlebu: Sounds good.

test/test.ts
scottbommarito: yes, since it's being resumed and not restarted no message will be sent.

test/template/www/js/scenarioSyncRestart30.js
scottbommarito: I put 30 seconds here because the Android emulator is really slow and I didn't want there to be any issues. The iOS tester is a lot faster and could deal with it but I wanted to pick a relatively safe value because I didn't want slowness in the Android emulator to impact anything.

test/test.ts
scottbommarito: correction: it's not waiting long enough before resuming, so it doesn't update and doesn't send any messages

test/projectManager.ts
scottbommarito: Yea I agree with you, I was following the file's convention, but moving it into IEmulatorManager is better.

test/test.ts
dlebu: I would explicitly add a new message when resuming in the application, using the [resume](https://cordova.apache.org/docs/en/5.4.0/cordova/events/events.resume.html) event. You have to assert something in a test case, otherwise there is no purpose in it.

test/template/www/js/scenarioSyncRestart30.js
dlebu: Let me know if you want me to help you set up HAXM for the Android emulator. It makes it blazing fast.

scottbommarito: I've been doing most of my development on my Windows and it doesn't support HAXM ugh...

test/test.ts
scottbommarito: Ok I agree

test/template/www/js/scenarioSyncRestart30.js
dlebu: You can use the VS emulator for android on Windows, which is a lot faster. I don't remember if I used it for running the tests, you might need to change the [server url](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/gulpfile.js#L145) to your machine's [localhost loopback](http://stackoverflow.com/questions/28262276/visual-studio-2015-ctp5-android-emulator-loopback-to-localhost/32660110#32660110)

scottbommarito: I might try that, thanks!

scottbommarito: Also I cut the delay time in half to 15 seconds, it still works fine haha

scottbommarito:
should be a lot better now!
@dlebu @geof90 @lostintangent

dlebu:
Looks good to me! Optionally, you can add an example to the `README.md` file showing how to use the new option.

Issue #57, opened 2016-01-25
Marshmallow on-demand permissions on cordova-android@5.1.0 


zwacky:
As soon as I'm about to install the OTA update, I'm asked to allow file storage permissions on the device. This is rather ugly and can turn users away from updating.

Probably not entirely a code-push, but maybe others stumble upon the same issue.

Can this storage permission be omitted with Marshmallow somehow? declaring in the AndroidManifest or something?

![screenshot_20160125-173154](https://cloud.githubusercontent.com/assets/1093032/12557493/55e5a048-c38b-11e5-9a77-3c501e0611b2.png)

lostintangent:
Yeah that is obviously not a good experience. Let us take a look into this and see what we can do to mitigate it. Stay tuned!

dlebu:
Thanks for reporting this issue. One of our plugin dependencies (the file transfer plugin) does add the write external storage to the Android manifest file, which under the new permission rules, is a dangerous permission. However, to my knowledge, we never use that permission since we [download the files to the application sandbox, not the phone external storage](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/remotePackage.ts#L79). So in theory, we should not see that permission dialog when installing the update.

Can you give me more details about your scenario?
- What kind of build are you running? Debug or release? Do I need to sign the apk?
- Were you on a device or emulator when reproducing the issue?

I tried to reproduce it on an emulator with a debug build unsuccessfully.

One quick way to fix this issue is to change your android target version to 22 like this:
```xml
<platform name="android">
        <preference name="android-minSdkVersion" value="YOUR_MIN_SDK_VERSION" />
        <preference name="android-targetSdkVersion" value="22" />
</platform>
```
Make sure your minimum sdk version is at most 22 as well. But this is not ideal and if this is a real issue we need to look into it.

zwacky:
tested on actual device (nexus 5), debug build
- using debug build

there's currently a bug in 

reproducable:
- if you use xwalk, make sure you use 15 or less (16+ problems with emulators)
-

zwacky:
popup comes:
- tested on actual device (nexus 5), debug build

popup doesn't come:
- emulator (nexus 5), debug build

zwacky:
setting the `android-targetSdkVersion` to 22 does fix it though.

scottbommarito:
[...]

PR #96, opened 2016-04-15
Test improvements and mandatory install mode

scottbommarito:
sorry for bundling these two together, mandatory install mode was so small I just started working on the next task without thinking lol

the general change here is that every test would call setupScenario and setupScenario would setupTemplate as well as doing what it does now. I also reworked a lot of code to make it flow better.


scottbommarito:
@geof90 @lostintangent @dlebu 
everything finally works and this is now ready for review

plugin.xml
```diff
@@ -7,8 +7,8 @@
         <repo>https://github.com/Microsoft/cordova-plugin-code-push.git</repo>
 
         <dependency id="code-push" version=">=1.8.0-beta" />
-        <dependency id="cordova-plugin-file-transfer" version=">=1.3.0" />
-        <dependency id="cordova-plugin-file" version=">=3.0.0" />
+        <dependency id="cordova-plugin-file" url="https://github.com/apache/cordova-plugin-file.git" commit="1e2593f42455aa78d7fff7400a834beb37a0683c" />
```
dlebu: Did you talk to the Cordova guys by any chance about this issue? When are they planning a new release of the file plugin?

test/projectManager.ts
```diff
@@ -113,28 +135,79 @@ export class ProjectManager {
...
 
     /**
      * Builds a specific platform of a Cordova project. 
      */
-    public static buildPlatform(projectFolder: string, targetPlatform: platform.IPlatform): Q.Promise<void> {
-        return ProjectManager.execAndLogChildProcess("cordova build " + targetPlatform.getCordovaName(), { cwd: projectFolder });
+    public static buildPlatform(projectFolder: string): Q.Promise<string> {
+        var targetPlatform = platform.PlatformResolver.resolvePlatform(tu.TestUtil.readTargetPlatform());
+        
+        // don't print output here because the iOS build process outputs so much nonsense that it buffer overflows and exits the entire test process
```
dlebu: An alternative here would have been increasing the [buffer size](https://github.com/Microsoft/cordova-plugin-code-push/pull/96/files#diff-8a098a7d94b3a7fe31827cde0cfeb7b6L190). However, I agree the build output is not useful during tests.

test/projectManager.ts
```diff
@@ -113,28 +135,79 @@ export class ProjectManager {
...
 
     /**
-     * Runs a project to a given target / platform.
+     * Launch the test app on the given target / platform.
+     */
+    public static launchApplication(namespace: string): Q.Promise<string> {
+        var targetPlatform = platform.PlatformResolver.resolvePlatform(tu.TestUtil.readTargetPlatform());
+        
+        var emulatorManager = targetPlatform.getEmulatorManager();
+        if (emulatorManager) {
```
dlebu: I would remove the check for `emulatorManager` in all these methods, now that we have an emulator manager for both platforms.

typings/codePush.d.ts
```diff
@@ -299,14 +299,19 @@ declare enum InstallMode {
...
     
     /**
      * If installMode === ON_NEXT_RESUME, the minimum amount of time (in seconds) which needs to pass with the app in the background before an update install occurs when the app is resumed.
      */
     minimumBackgroundDuration?: number;
+    
+    /**
+     * Used to specify the InstallMode used for the install operation if the update is mandatory. This is optional and defaults to InstallMode.IMMEDIATE.
+     */
+    mandatoryInstallMode?: InstallMode;
```
dlebu: I wonder if using the other `installMode` value as the default is not a better solution here. This way, we would preserve backwards compatibility with developers already using the plugin but not using the new `mandatoryInstallMode` option. @lostintangent

dlebu:
Please update the readme file as well with the new option and optionally add or modify an existing code sample.

test/platform.ts
```diff
@@ -151,68 +161,102 @@ export class IOSEmulatorManager implements IEmulatorManager {
...
             .then<string>(applicationLabel => {
-                return tu.TestUtil.getProcessOutput("xcrun simctl spawn booted launchctl stop " + applicationLabel, undefined, true);
-            })
-            .then<void>(output => {
-                console.log(output);
+                // kill the app if we found the process
+                return tu.TestUtil.getProcessOutput("xcrun simctl spawn booted launchctl stop " + applicationLabel, undefined);
+            }, (error) => {
+                // we couldn't find the app's process so it must not be running
+                return Q.resolve(error);
```
geof90: Why are we resolving with an error? Should we reject instead?

typings/codePush.d.ts
scottbommarito: Just spoke to @lostintangent, this is what he wants. He was under the impression that mandatory updates were already immediate, but that something must have changed to remove that bug/feature lol

plugin.xml
scottbommarito: @silhouettes asked them about 5 days ago and they have not yet responded

test/projectManager.ts
scottbommarito: didn't even notice that haha, good catch! i'm ok with including android build output but the iOS output is absolutely horridly excessive

test/projectManager.ts
```diff
@@ -181,25 +257,25 @@ export class ProjectManager {
     }
 
     /**
-     * Executes a child process and logs its output to the console.
+     * Executes a child process and logs its output to the console and returns its output in the promise as a string
      */
-    public static execAndLogChildProcess(command: string, options?: child_process.IExecOptions): Q.Promise<void> {
-        var deferred = Q.defer<void>();
+    public static execAndLogChildProcess(command: string, options?: child_process.IExecOptions, output: boolean = true): Q.Promise<string> {
```
geof90: If the logging behaviour is now toggle-able, would it make sense to just name this `execChildProcess`?

test/projectManager.ts
geof90: "Don't log the iOS build output because is too verbose and overflows the buffer." (being slightly nit-ish here about code comments since this is open source code)

test/projectManager.ts
```diff
@@ -113,28 +135,79 @@ export class ProjectManager {
...
+    /**
+     * Kill the test app on the given target / platform.
      */
-    public static runPlatform(projectFolder: string, targetPlatform: platform.IPlatform, skipBuild: boolean = false, target?: string): Q.Promise<void> {
+    public static endRunningApplication(namespace: string): Q.Promise<string> {
+        var targetPlatform = platform.PlatformResolver.resolvePlatform(tu.TestUtil.readTargetPlatform());
+        
+        var emulatorManager = targetPlatform.getEmulatorManager();
+        if (emulatorManager) {
+            return emulatorManager.endRunningApplication(namespace);
```
geof90: Rename this to `appNamespace`? `namespace` is a keyword in TypeScript.

test/platform.ts
scottbommarito: well, this method is used not just to kill the app but also to guarantee an app isn't running. when the app is already gone, then the result state--the app is not running--is the same, so it's not necessary to reject because the desired end state of the promise has been maintained.

test/projectManager.ts
geof90: Also, `logOutput` may be a better name instead of `output`.

plugin.xml
dlebu: They [just replied in the issue thread](https://github.com/Microsoft/cordova-plugin-code-push/issues/57#issuecomment-211504282). They will be releasing soon - the release is already listed on github, but not yet on npm. I think we should wait for the new version.

test/projectManager.ts
scottbommarito: ahhh I was wondering why the naming wasn't consistent everywhere

test/platform.ts
geof90: Ah i get it now. 👍

plugin.xml
scottbommarito: awesome! will do

geof90:
LGTM!

README.md
```diff
@@ -263,7 +263,7 @@ codePush.notifyApplicationReady(notifySucceeded?, notifyFailed?);
 
 Notifies the CodePush runtime that a freshly installed update should be considered successful, and therefore, an automatic client-side rollback isn't necessary. It is mandatory to call this function somewhere in the code of the updated bundle. Otherwise, when the app next restarts, the CodePush runtime will assume that the installed update has failed and roll back to the previous version. This behavior exists to help ensure that your end users aren't blocked by a broken update.
 
-If you are using the `sync` function, and doing your update check on app start, then you don't need to manually call `notifyApplicationReady` since `sync` will call it for you. This behavior exists due to the assumption that the point at which `sync` is called in your app represents a good approximation of a successful startup.
+If you are using the `sync` function, and doing your update check on app start, then you don't need to manually call `Ready` since `sync` will call it for you. This behavior exists due to the assumption that the point at which `sync` is called in your app represents a good approximation of a successful startup.
```
dlebu: Is this change intentional?

scottbommarito: omg ugh nope

dlebu:
LTGM! Don't forget to remove the file plugin dependency change from plugin.xml.

scottbommarito:
NOTE: i'm waiting on a release of the cordova file plugin before I merge this, if there isn't a release by tomorrow night i'll probably just merge anyway but given that they said it was soon i'm going to wait a little

dlebu:
@scottbommarito @geof90 Cordova customers are used to getting Cordova plugins directly from the GitHub URL, that's the reason for which we check in .js sources as well. Going forward, when implementing new features, please keep master clean until after you released to npm. This means multiple things:

- Creating a new **base** branch and bumping up the version whenever you take on new work.
- Every time we code a new feature, branch the **base** branch, and create PRs against it. You will probably have more than one PRs between npm releases, and this will keep master clean.
- Once you have all your PRs in, create a new release from that branch, and publish to npm.
- Finally, merge your **base** branch to **master**.

geof90:
@dlebu I am apprehensive that we should do this, for the following reasons:

1. It seems common practice for master to be ahead of NPM releases. The cordova file plugin, as we have already experienced, is an example, their marshmellow permissions fix is on master but not on NPM. The react native project is also an example where they explicitly state in their CONTRIBUTING.md that master is unsafe.

2. Having latest changes on master lets users who want to be on the bleeding edge use the latest features. It also makes it easier for the community to send PRs, unless we document this other practice you proposed in the CONTRIBUTING.md

3. This works best in practice if every PR that is merged to master is well tested, and master is itself periodically tested. We are right now already in the midst of getting CI setup for this project, which will aid us in this.

Overall, i prefer that we keep our current practice in order to maintain our agility, but i am open to discussion. What do the rest think?

dlebu:
@geof90

1. I agree it is common practice, but I am mostly concerned about Visual Studio users, where until recently, there was no way to add a third party plugin by name, only by URL. Keeping things separated also enables you to do patches if you discover critical bugs while working on new features. 
2. They could still do this by specifying the branch name in the URL.
3. Agreed.
4. Product quality and breaking changes were my main concerns. The CI server offers a solution for quality, but not for API breaking changes. One scenario I had in mind is documentation, which can get out of date when coding new features. Or, the doc is up to date with master but the user installed the current version of the plugin from npm, which is out of date with the documentation. Most people look at the documentation on github, even if they installed from npm.

That said, if CI integration happens keeping master as base could still be an option if we are fine with having the documentation out of date. Personally, I think that keeping the documentation and samples up to date is a key factor in engaging developers with the product.

silhouettes:
Is it common for Cordova users to just pull the latest on master, as opposed to locking down a specific commit or release tag? (And looking up documentation by release tag?)

The former (which we seem to be trying to engineer around here) seems like bad practice in general, but I don't know if that is the norm.

dlebu:
@silhouettes In the Visual Studio documentation [we tell them](http://taco.visualstudio.com/en-us/docs/manage-plugins/#Custom) to use the repo URL. Not sure about developers that do not use Visual Studio.

scottbommarito:
@dlebu IMO this is at its core a Visual Studio problem. The common convention is that master represents the most updated version, and that all changes are merged into it, and then, when master is at a stable place, a "snapshot" of it is taken and put somewhere, in this case npm. Because visual studio doesn't have access to where we put our releases it has to choose the next best thing, which is, in this case, our unstable master branch. What really needs to happen is that Visual Studio needs to add support for npm and then the docs in both Visual Studio and our project need to reflect it, not that we need to reorganize our branches.

dlebu:
@geof90 @silhouettes @scottbommarito I am fine with whatever you guys decide, given that the quality will be gated by the CI server.

silhouettes:
Okay, so given that this is the way that older versions of Visual Studio behave and we can't change it, maybe it makes sense to have a `feature`/`dev`/`next` branch that we merge all of our changes into.

In terms of the tradeoff, I think it won't cost us too much in process, and it helps us be extra-defensive against the issues Daniel mentioned (even if they could be argued to be user error). This might also be good practice in general, for example in scenarios where we want to ship a hotfix, but there are other features that have been pushed which we do not want to ship yet. Otherwise, combined with the VS thing, we would have the pressure of having every merged PR be release-ready, which is a requirement that we really don't need to have looming over us.

scottbommarito:
@silhouettes what if that whenever we released, we merged all the changes into a branch, and had in our docs for people using Visual Studio to use that branch instead of master? that way master is the latest and greatest while VS also has a URL to pull from.

PR #97, opened 2016-04-16
reportStatus changes

geof90:
This PR makes some changes to the reportStatus mechanism.

1. Status reports are sent immediately, instead of being queued up to be sent when the app goes into the background.

2. Sends the previous deployment key and label, so that the server does not have to save every version every client ID is on. That also allows the server to decrement the active users from the old deployment key and increment the active users for the new deployment key, effectively recording an update from one deployment key to another.


lostintangent:
LGTM!

ryuyu:
Lgtm

PR #98, opened 2016-04-20
Fix build warning

lostintangent:
This PR simply fixes a build warning on iOS. It doesn't impact the app at all, but it may cause concern for users and it introduces unnecessary noise in the build output.

screen shot 2016-04-20 at 4 02 53 pm


dtivel:
Looks good.

PR #101, opened 2016-04-22
Automated testing

scottbommarito:
-can now run multiple platforms in a single test
-now starts/restarts emulators whenever you run a task
-every possible configuration of platforms possible in a single command
-slightly improved test output
-added more command line flags and options to the mocha command used to run tests

NOTE: I took out the "application_resumed" test message. Sometimes, cordova would just not send the event leading to unpredictable failures that would not reflect whether or not the plugin was actually working. We know the operation we are performing is a resume, and if the app sends the rest of the messages we can know that it's running correctly anyway.


scottbommarito:
@geof90 @dlebu @lostintangent 

i hope u love gigantic prs hehe

CONTRIBUTING.md
```diff
@@ -38,17 +54,78 @@ OS            | Supported tests
 OS X          | Android, iOS
 Windows       | Android
 
+### Test descriptions
+
+The tests first build the app.
+
+They then check if the required emulators are currently running. If running only Android tests, it checks for a running Android emulator, and if running only iOS tests, it checks for a running iOS simulator. If running both, it will check for both.
```
geof90: The first sentence seems sufficient. Probably can remove from `If running only Android tests ... it will check for both.`

CONTRIBUTING.md
```diff
@@ -38,17 +54,78 @@ OS            | Supported tests
...
+If they are not, then it attempts to start an Android emulator named ```emulator``` (if one exists) and an iOS simulator named ```iPhone 6s (9.3)```. If these emulators don't exist on your computer, the task will exit and the tests will fail. 
+If you would like the tests to always restart the necessary emulators, add a ```-clean``` to the end of the command you'd like to run.
 
-##### iOS
+The desired unit tests are then run.
+There is a both a full unit test suite and a "core" set of unit tests that you may run. If you would like to run only the core tests, add a ```-core``` to the end of the command you'd like to run.
+If you would like to pull the plugin from NPM rather than running the tests on the local version, add a ```-npm``` to the end of the command you'd like to run.
+If you would like to skip building and checking for emulators, add a ```-fast``` to the end of the command you'd like to run.
```
geof90: Note that all of these are not line separated, you need an empty line in order for the individual paragraphs to be separated in GitHub Markdown

CONTRIBUTING.md
scottbommarito: that makes sense LOL

CONTRIBUTING.md
```diff
@@ -38,17 +54,78 @@ OS            | Supported tests
...
+
+To run all of the unit tests on Android:
 ```
 gulp test-android
 ```
+
+#### More examples
+
+All possible testing configurations have tasks!
+The flags should be ordered as following: android, ios, uiwebview, wkwebview, core, npm, fast, clean
```
geof90: as follows

gulpfile.js
```diff
@@ -8,6 +8,12 @@ var testPath = "./test";
 var binPath = "./bin";
 var tsFiles = "/**/*.ts";
 
+var androidEmulatorName = "emulator";
+var iOSEmulatorName = "iPhone 6s Plus (9.3) [";
```
geof90: Is the ` [` intentional?

scottbommarito: yes, because there is a "iPhone 6s Plus (9.3) with Apple Watch" simulator as well so the "[" has to be there to specify that it's just the "iPhone 6s Plus (9.3)" emulator

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
-});
+function startEmulators(callback, restartIfRunning, android, ios) {
+    // called when an emulator is initialized successfully
+    var emulatorsInit = 0;
+    function onEmulatorInit(emulator) {
+        ++emulatorsInit;
+        console.log(emulator + " emulator is ready!");
+        if (emulatorsInit === ((android ? 1 : 0) + (ios ? 1 : 0))) {
+            console.log("All emulators are ready!");
+            callback(undefined);
```
geof90: does simply calling `callback();` work?

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+        }
+    }
+    
+    // called to check if an Android emulator is initialized
+    function androidEmulatorReady(onFailure) {
+        console.log("Checking if Android emulator is ready yet...");
+        // dummy command that succeeds if emulator is ready and fails otherwise
+        spawnCommand("adb", ["shell", "pm", "list", "packages"], (code) => {
```
geof90: Lets stick to ES5 syntax (no arrow functions) for this gulpfile since 1. It's the convention throughout this file and 2. It won't work on older versions of node.

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+                console.log("Android emulator is not ready yet!");
+                return onFailure();
+            }
+        }, true);
+    }
+    // called to check if an iOS emulator is initialized
+    function iOSEmulatorReady(onFailure) {
+        console.log("Checking if iOS emulator is ready yet...");
+        // dummy command that succeeds if emulator is ready and fails otherwise
+        spawnCommand("xcrun", ["simctl", "getenv", "booted", "asdf"], (code) => {
```
geof90: `function(code) { ...`

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+            if (!code) return onEmulatorInit("iOS");
+            else {
+                console.log("iOS emulator is not ready yet!");
+                return onFailure();
+            }
+        }, true);
+    }
+    // kills the Android emulator then starts it
+    function killThenStartAndroid() {
+        spawnCommand("adb", ["emu", "kill"], () => {
```
geof90: function() { ...

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+    } else {
+        if (android) killThenStartAndroid();
+        if (ios) killThenStartIOS();
+    }
+    // This needs to be done so that the task will exit.
+    // The command that creates the Android emulator persists with the life of the emulator and hangs this process unless we force it to quit.
+    gulp.doneCallback = (err) => {
+        process.exit(err ? 1 : 0);
+    }
+}
```
geof90: Seems like collapsing `startEmulators(callback, restartIfRunning, platforms) {... ,killThenStart(platform ...), emulatorReady(platform ...)` etc together would shorten the code here because they pretty much do the same thing with some small differences (command args, killing android process).

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+function getEmulatorTaskNameSuffix(android, ios) {
+    var emulatorTaskNameSuffix = "";
+    
+    if (android) emulatorTaskNameSuffix += "-android";
+    if (ios) emulatorTaskNameSuffix += "-ios";
+    // "emulator" instead of "emulator-android-ios"
+    if (android && ios) emulatorTaskNameSuffix = "";
+    
+    return emulatorTaskNameSuffix
+}
```
geof90: more simply:

```
function getEmulatorTaskNameSuffix(android, ios) {
    if (!!android == !!ios) return "";
    else if (android) return "-android";
    else return "-ios";
}
```

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
 
-gulp.task("test", function (callback) {
-    runSequence("default", "test-android", "test-ios", callback);
-});
\ No newline at end of file
+var emulatorTaskNamePrefix = "emulator";
+for (var android = 0; android < 2; android++) {
+    for (var ios = 0; ios < 4; ios++) {
+        generateEmulatorTasks(emulatorTaskNamePrefix + getEmulatorTaskNameSuffix(android, ios), android, ios);
```
geof90: I'm really struggling to understand this.
android = 0, ios = 0
android = 1, ios = 1
android = 1, ios = 2
android = 1, ios = 3
All result in the same task name?

geof90: There are really only two emulators, unlike the tests where there are multiple config's per platform (UIWebView, WKWebView) so let's simplify this, there's no need for a for loop here.

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+                
+                taskName += taskNameSuffix;
+                
+                var options = {};
+                if (android) options.android = true;
+                if (ios) options.ios = true;
+                if (ios % 2 === 1) options.uiwebview = true;
+                if (ios >= 2) options.wkwebview = true;
+                if (core) options.core = true;
+                if (npm) options.npm = true;
```
geof90: If you changed it to the following it would be cleaner IMO:

```javascript
var taskName = taskNamePrefix;
var options = {};

if (android) {
    taskNameSuffix += "-android";
    options.android = true;
}

if (ios) {
    taskNameSuffix += "-ios"
    options.ios = true;
    options.uiwebview = ios % 2 == 1;
    options.wkwebview = ios >= 2;
    if (options.uiwebview && !options.wkwebview) {
        taskNameSuffix += "-uiwebview";
    } else if (options.wkwebview && !options.uiwebview) {
        taskNameSuffix += "-wkwebview";
    }
}

if (android && ios ===3) taskNameSuffix = "";

if (core) {
    taskNameSuffix += "-core";
    options.core = true;
}

if (npm) {
    taskNameSuffix += "-npm";
    options.npm = true;
}

generateTestTasks(taskName, options);
```

CONTRIBUTING.md
```diff
@@ -11,6 +11,11 @@ Instead, modify the sources in the ```www``` folder and compile a new version of
 
 ```node.js``` and ```npm``` are needed for building this project. ```npm``` comes bundled with the ```node.js``` installer. You can download the ```node.js``` installer here: https://nodejs.org/download/. 
 
+To run Android tests, make sure you have ```sdk\tools``` and  ```sdk\platform-tools``` in your PATH.
+
+If you would like the tests to start the emulators for you, configure an Android emulator using the AVD and name it ```emulator```.
```
dlebu: This should not be hard coded. Can you make it a command line parameter to gulp?

CONTRIBUTING.md
```diff
@@ -29,7 +34,18 @@ Follow these steps to build a new version of the plugin:
...
 	This will compile the sources and place them in the ```bin``` folder. Any compilation errors will be displayed in the console.
 
-### Test
+## Test
+
+### Environment setup
+
+Make sure you can build the plugin by following the steps above.
+
+To run Android tests, make sure you have ```sdk\tools``` and  ```sdk\platform-tools``` in your PATH.
```
dlebu: Duplicated information? The above paragraph states the same thing.

CONTRIBUTING.md
```diff
@@ -38,17 +54,85 @@ OS            | Supported tests
 OS X          | Android, iOS
 Windows       | Android
 
+### Test descriptions
+
+The tests first build the app.
+
+They then check if the required emulators are currently running.
+
+If they are not, then it attempts to start an Android emulator named ```emulator``` (if one exists) and an iOS simulator named ```iPhone 6s (9.3)```. If these emulators don't exist on your computer, the task will exit and the tests will fail.
```
dlebu: Same for iOS, we should not hard code the simulator names. How about we pick the first available emulator/simulator if the user did not specify one already?

CONTRIBUTING.md
```diff
@@ -38,17 +54,85 @@ OS            | Supported tests
...
+
+If you would like the tests to always restart the necessary emulators, add a ```-clean``` to the end of the command you'd like to run.
 
-##### iOS
+The desired unit tests are then run.
 
-To run the iOS tests:
+There is a both a full unit test suite and a "core" set of unit tests that you may run. If you would like to run only the core tests, add a ```-core``` to the end of the command you'd like to run.
```
dlebu: Why do we need to split the test suite? We have pretty good coverage and running time is under 10 minutes.

CONTRIBUTING.md
```diff
@@ -38,17 +54,85 @@ OS            | Supported tests
...
+#### Android
+
+To run all of the unit tests on Android:
 ```
 gulp test-android
 ```
+
+#### More examples
+
+All possible testing configurations have tasks!
```
dlebu: Why the exclamation mark?

gulpfile.js
dlebu: Again, we should not hardcode these values, but pick one of the already installed emulator/simulator on the machine.

gulpfile.js
```diff
@@ -29,22 +35,64 @@ var tsCompileOptions = {
...
+        if (!silent) console.log("" + data);
+    });
+
+    process.stderr.on('data', function (data) {
+        if (!silent) console.error("" + data);
     });
     
-    process.on('exit', function(code) {
-        callback(code === 0 ? undefined : "Error code: " + code);
+    process.on('exit', function (code) {
```
dlebu: I think I was missing a `process.on('error', ... )` handler here.

gulpfile.js
geof90: Basically the idea is to do number -> boolean conversion once instead of checking numbers everywhere, which can be brittle.

CONTRIBUTING.md
scottbommarito: why not...lol

gulpfile.js
scottbommarito: yea I think I just forgot to clean this code up, I can just reduce the < 4 to a < 2 and I think it'll work pretty much off the bat

gulpfile.js
scottbommarito: I have not seen this `!!x` syntax before, will ask u

CONTRIBUTING.md
scottbommarito: Each platform takes ten minutes...it takes 30-40 minutes total for all of them (android, ios-ui, ios-wk). Since we want CI that runs in about 10-15 minutes we had to prune the tests, and then run the full suite nightly.

gulpfile.js
geof90: I really think you should just get rid of the loops, there's no need for nested loops when there are really only 2 tests.

CONTRIBUTING.md
scottbommarito: first available simulator would be really old since iOS installs a ton of simulators when u install xcode. i'll rewrite this code to grab the newest.

test/platform.ts
```diff
@@ -269,14 +307,33 @@ export class PlatformResolver {
...
+    public static resolvePlatforms(cordovaPlatformNames: string[]): IPlatform[] {
+        var platforms: IPlatform[] = [];
+
+        for (var i = 0; i < cordovaPlatformNames.length; i++) {
+            var resolvedPlatform: IPlatform = PlatformResolver.resolvePlatform(cordovaPlatformNames[i]);
+            if (resolvedPlatform) platforms.push(resolvedPlatform);
+            else {
+                // we could not find this platform in the list of platforms, so abort
+                console.error("Unsupported platform: " + cordovaPlatformNames[i]);
+                return undefined;
```
geof90: Just `return` will be sufficient.

CONTRIBUTING.md
scottbommarito: wasn't sure if gulp supported command line arguments lol...in fact honestly if we can use command line arguments I can rewrite half of the gulp file oh god

CONTRIBUTING.md
scottbommarito: actually i forgot to remove this in the plugin env setup, since this stuff is exclusive to running tests and not for building

CONTRIBUTING.md
scottbommarito: jk, just talked with Geoffrey, gonna leave the enumeration code in because the union rules (android + ios-ui + ios-wk = test, ios-ui + ios-wk = ios, etc) makes supplying what tests to run kinda impractical, but emulator can be passed by command line

geof90: I think emulator names should be passed by command line with defaults if not provided, like what Daniel suggests. As to the test task names (android, ios+ui, ios+wk etc..) I will leave it up to you since it looks like its going to be a huge undertaking to refactor everything

test/projectManager.ts
```diff
@@ -136,57 +131,72 @@ export class ProjectManager {
...
     /**
      * Builds a specific platform of a Cordova project. 
      */
-    public static buildPlatform(projectFolder: string): Q.Promise<string> {
-        var targetPlatform = platform.PlatformResolver.resolvePlatform(tu.TestUtil.readTargetPlatform());
-        
-        // don't log the iOS build output because it is too verbose and overflows the buffer
+    public static buildPlatform(projectFolder: string, targetPlatform: platform.IPlatform): Q.Promise<string> {
+        console.log("Building " + targetPlatform.getCordovaName() + " project in " + projectFolder);
+        // don't print output here because the iOS build process outputs so much nonsense that it buffer overflows and exits the entire test process
```
geof90: Why did you change the comment?

test/testUtil.ts
```diff
@@ -9,45 +9,94 @@ import path = require("path");
...
-
-    public static MOCK_SERVER_OPTION_NAME: string = "--mockserver";
-    public static PLATFORM_OPTION_NAME: string = "--platform";
-    public static TARGET_OPTION_NAME: string = "--target";
+    public static ANDROID_PLATFORM_OPTION_NAME: string = "--android";
+    public static IOS_PLATFORM_OPTION_NAME: string = "--ios";
     public static SHOULD_USE_WKWEBVIEW: string = "--use-wkwebview";
+    public static TEST_RUN_DIRECTORY: string = "--test-directory";
+    public static TEST_UPDATES_DIRECTORY: string = "--updates-directory";
+    public static CORE_TESTS_ONLY: string = "--core-tests";
```
geof90: Just --core, to match what you had earlier?

test/testUtil.ts
```diff
@@ -9,45 +9,94 @@ import path = require("path");
...
+     */
+    public static readShouldUseWkWebView(): number {
         var shouldUseWkWebView = TestUtil.readMochaCommandLineOption(TestUtil.SHOULD_USE_WKWEBVIEW);
-        return shouldUseWkWebView ? JSON.parse(shouldUseWkWebView) : false;
+        switch (shouldUseWkWebView) {
+            case "true":
+                console.log("WkWebview");
+                return 1;
+            case "both":
+                console.log("Both WkWebview and UIWebView");
```
geof90: Nit: "WKWebView"

gulpfile.js
scottbommarito: did it like this cuz it was easy after already doing all that code haha but yea i'll fix

test/projectManager.ts
scottbommarito: woopsidaisy

test/testUtil.ts
scottbommarito: woopsidaisy

CONTRIBUTING.md
scottbommarito: i will keep the task names as is just because i think it works better for actual execution (ie "test" is shorthand for "test --android --ios-ui --ios-wk") but emulator names passed by command line sounds groovy

gulpfile.js
scottbommarito: actually i think it's better looped like this, there are 6 tasks getting created and taking out the loop structure is much uglier than what's here

gulpfile.js
scottbommarito: i raise you one
`return !!android ? (!!ios ? "" : "-android") : (!!ios ? "-ios" : "");`

CONTRIBUTING.md
dlebu: Can we make them run in parallel? I think the only thing stopping us from doing that is the test project paths, which are common, but we should be able to change that. It's not necessary to do it now, but it would be nice to have.

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
\ No newline at end of file
+var emulatorTaskNamePrefix = "emulator";
+for (var android = 0; android < 2; android++) {
+    for (var ios = 0; ios < 4; ios++) {
+        generateEmulatorTasks(emulatorTaskNamePrefix + getEmulatorTaskNameSuffix(android, ios), android, ios);
+    }
+}
+                
+function generateTestTasks(taskName, options) {
+    gulp.task(taskName + "-fast", function (callback) {
```
dlebu: For clarity, I think it's best to keep gulp tasks declared as globals and not generated at runtime.

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+        runSequence("default", emulatorTaskName, taskName + fastSuffix, callback);
+    });
+    
+    gulp.task(taskName + "-clean", function (callback) {
+        runSequence("default", emulatorTaskName + cleanSuffix, taskName + fastSuffix, callback);
+    });
+}
+
+// procedurally generate tasks for every possible testing configuration
+var taskNamePrefix = "test";
```
dlebu: How is this simpler than the gulp tasks we had before? I think this would get hard to maintain for any engineer that did not write the actual code. I would have one gulp task for every flavor, similar to what we had before, and compose those into more complex tasks using gulp task dependencies.

gulpfile.js
scottbommarito: does that not go through this with a truthy code?

dlebu: It may not always go through the exit handler: https://nodejs.org/api/child_process.html#child_process_event_error

gulpfile.js
```diff
@@ -124,28 +172,197 @@ gulp.task("default", function (callback) {
...
+var cleanSuffix = "-clean";
+var fastSuffix = "-fast";
 
-gulp.task("test-ios-wkwebview", function (callback) {
-    var command = "mocha";
-    var args = ["./bin/test", "--mockserver", "http://127.0.0.1:3000", "--platform", "ios", "--use-wkwebview", "true"];
-    executeCommand(command, args, callback);
-});
+// generate tasks for starting emulators
+function generateEmulatorTasks(taskName, android, ios) {
```
dlebu: I feel like this belongs to the test set-up process and not to the gulp file. We should really aim at keeping the build file (gulp file) simple and easy to understand. You could add a new parameter that fires up the emulator if you really need to control this from a gulp task, but I think the emulators should be started by default if they are not running.

dlebu: I just read my comment again and it is not clear. By `this` I meant emulator operations, like stopping and starting them. Sorry for the confusion.

gulpfile.js
dlebu: @geof90's version looks better, but I am trying to understand why do we need a gulp task for every possible combination of options. If we want to keep things simple, we could have gulp support for the default ones (say check-in run and nightly run), and the rest should be fine with support via mocha.

dlebu: We are not going to use every possible gulp task anyway.

test/platform.ts
```diff
@@ -62,6 +67,11 @@ export interface IEmulatorManager {
      * Prepares the emulator for a test.
      */
     prepareEmulatorForTest(appId: string): Q.Promise<string>;
```
dlebu: The platform interface would be a good place for the simulator/emulator related operations.

test/testUtil.ts
```diff
@@ -9,45 +9,94 @@ import path = require("path");
...
+    public static ANDROID_PLATFORM_OPTION_NAME: string = "--android";
+    public static IOS_PLATFORM_OPTION_NAME: string = "--ios";
     public static SHOULD_USE_WKWEBVIEW: string = "--use-wkwebview";
+    public static TEST_RUN_DIRECTORY: string = "--test-directory";
+    public static TEST_UPDATES_DIRECTORY: string = "--updates-directory";
+    public static CORE_TESTS_ONLY: string = "--core-tests";
+    public static PULL_FROM_NPM: string = "--npm";
+    
+    public static IOSServerUrl = "http://127.0.0.1:3000";
+    public static AndroidServerUrl = "http://10.0.2.2:3000";
```
dlebu: For Android, this is not always the case. In the case of Visual Studio Android Emulator the localhost loopback for the emulator is different, that's why I believe keeping it as a mocha command line parameter is a good idea.

CONTRIBUTING.md
scottbommarito: i briefly looked into parallel tests, the issue that i had was the server port being the same. if we can change the server ports without issue then it should work in parallel easily

gulpfile.js
scottbommarito: this is for automated testing and must be in the gulp file

test/testUtil.ts
scottbommarito: I really like the mocha commands being simpler. that said, I can put back the server url flag while keeping this the default. the gulp test before would always pass that server url for android so I moved it here

gulpfile.js
scottbommarito: also i think that having it in the testing process is honestly nicer. if you accidentally started tests without having your emulator ready it automatically starts them. on the other hand, if your emulator is already running it has basically zero impact on your runtime.

gulpfile.js
scottbommarito: it was not meant to be simpler--it was meant to be more extensive. currently it's not possible to make the complex tasks through dependencies as creating the test directory is not in its own task and is run at the beginning of any set of tests. i think giving creating the test directory its own task has potential but it could be a little annoying to implement

gulpfile.js
scottbommarito: ironically this may explain a weird issue i saw before and was able to get around but with some confusion

CONTRIBUTING.md
dlebu: Although supported, gulp command line parameters are a bit odd to use. If this is the first scenario where we need them, you could add the parameters to `mocha`, not `gulp` instead. That, coupled with moving the emulator managing code from the gulp file to the tests makes more sense.

test/platform.ts
scottbommarito: but isn't the emulator manager the best place for simulator/emulator related operations

gulpfile.js
dlebu: What I meant is moving [these kind of functions](https://github.com/Microsoft/cordova-plugin-code-push/pull/101/files#diff-b9e12334e9eafd8341a6107dd98510c9R233) directly into the test setup. That way, the gulp file is not responsible with starting the emulators any more.

CONTRIBUTING.md
scottbommarito: (the duplicate in the plugin env setup has been removed)

gulpfile.js
```diff
@@ -124,28 +199,203 @@ gulp.task("default", function (callback) {
...
+        runSequence("default", emulatorTaskName + cleanSuffix, taskName + fastSuffix, callback);
+    });
+}
+
+// procedurally generate tasks for every possible testing configuration
+var taskNamePrefix = "test";
+for (var android = 0; android < 2; android++) {
+    // 0 = don't run android tests
+    // 1 = run android tests
+    for (var ios = 0; ios < 4; ios++) {
```
dlebu: Don't forget that gulp is supposed to be a build system. Think of it as the equivalent of a `.csproj` file in Visual Studio. The fact that we can implement everything using gulp tasks doesn't mean we should. In my opinion, keeping the build file lightweight and moving the heavy code to the test sources makes sense. If we have a lot of options for tests and test filtering, mocha supports all of that out of the box. We don't need to reinvent the wheel in gulp.

Try to think about responsibilities. Is the build system supposed to perform setup for tests? Or is it the tests that need to handle that?

test/platform.ts
dlebu: That is correct, it is. Ignore my comment.

dlebu: The emulator manager is also the right place for [this](https://github.com/Microsoft/cordova-plugin-code-push/pull/101/files#diff-b9e12334e9eafd8341a6107dd98510c9R206).

gulpfile.js
scottbommarito: ultimately, if we're just moving code from the build system to the tests and vice versa, we're not getting rid of complexity--we're just moving it around. i could move the emulator starting code into test.js but ultimately that would just be adding more complexity into the already gigantic test.js file and I don't think it would really accomplish much.

a build system includes running tests. i don't see having the test setup here any problem. i think the issue is that it's not immediately evident what the tasks are, not that the file is too big or too hard to understand what's going on. although i argue that if you're really confused about what the tasks are you should just read the contributing.md which i believe does a fairly sufficient job, i agree that there should be more immediate transparency as to what the tasks are just by looking at the gulpfile.js itself, but i don't think the emulator startup being in the gulpfile is an issue

dlebu: We can always split the `test.js` file into multiple files so that it is more maintainable. That is not enough reason to keep test initialization code in the gulp file. A build system does include kicking off test runs, but not performing test set-up or clean-up. That responsibility is of the test itself, not of the build system.

I think you guys should meet and talk about the design of the tests. If everyone agrees with this approach, I am fine with it, but I personally find it unnecessary complicated.

@dtivel @silhouettes @lostintangent Thoughts?

scottbommarito: well I mean I do agree that the OPTIMAL place for the emulator initialization is in the tests, however it seems like any time I change something structurally in the tests it breaks EVERYTHING and then I have to spend days trying to fix it. right now i would like to move onto getting the react-native tests working which is going to involve a lot of the same framework, and then i can perform some optimizations on both of them, and think having the emulator initialization in the gulpfile is less priority

dlebu: That means that we not only have a design issue, but also a stability issue and we should fix it. This argument is even stronger given that you mentioned you want to use the same framework for react native. If the test framework or the tests themselves are not stable, enabling continuous integration will not have benefits. It will only make things worse, and waste developer time on test failure investigations. My understanding was that the #96 refactor was intended to make the tests [flow better](https://github.com/Microsoft/cordova-plugin-code-push/pull/96).

scottbommarito: it's a step by step process. just like last PR made the flow better, this PR continues to make the flow better while adding automation support. the next PR will continue to improve the flow while also adding react-native support. the tests are stable. what is not stable is making breaking changes to the flow and expecting it to work easily. this is just another stable point in the process in which i think we should merge before they break until i finish the next improvement.

scottbommarito:
gulp tasks are now much cleaner

emulator setup is still in the gulpfile but will move to emulator manager tomorrow

test/platform.ts
scottbommarito: it is now in the emulator manager

scottbommarito:
@geof90 @dlebu @lostintangent

ok should be good now

CONTRIBUTING.md
```diff
@@ -38,17 +46,89 @@ OS            | Supported tests
...
 
+### Test descriptions
+
+The tests first build the app.
+
+They then check if the required emulators are currently running.
+
+If an Android emulator is not running, it attempts to boot an Android emulator named ```emulator```. You can specify an emulator by adding ```--androidemu yourEmulatorNameHere``` as a command line option to the gulp task.
+
+If an iOS simulator is not running, it attempts to boot the latest iOS iPhone simulator. You can specify a simulator by adding ```--iosemu yourSimulatorNameHere``` as a command line option to the gulp task.
```
dlebu: Given that the iOS platform uses simulators and not emulators, would it be a good idea to rename this option to something like `--iossim`? If we do rename it, we would lose consistency with the android option. If consistency is a concern, we can pick a name that can be used for both, like `--target`

test/template/config.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-    <widget id="io.cordova.test.codepush" version="1.5.0-beta" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
+    <widget id="com.microsoft.codepush.test" version="1.5.0-beta" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
```
dlebu: It would be nice to try to keep the test template version in sync with the plugin version.

dlebu:
@scottbommarito I had two minor comments, otherwise LGTM! Thanks for making the changes now, I believe it is a good dev time investment that will pay off in the long run.

CONTRIBUTING.md
scottbommarito: can't choose target for both because you can run both android and ios tests with one command. I think we might as well leave it like this for consistency and also because the emulator boot script calls the iOS simulator an emulator (so I could reuse the code easily)

gulpfile.js
```diff
@@ -29,22 +34,125 @@ var tsCompileOptions = {
...
+
+    process.stdout.on('data', function (data) {
+        if (!silent) console.log("" + data);
     });
+
+    process.stderr.on('data', function (data) {
+        if (!silent) console.error("" + data);
+    });
+    
+    process.on('error', function (code) {
```
geof90: [The callback arg should be an `error` object.](https://nodejs.org/api/child_process.html#child_process_event_error)

scottbommarito: gotta fix that in the gulpfile as well, thanks

scottbommarito: wait jk this is the gulpfile kappa

gulpfile.js
```diff
@@ -29,22 +34,125 @@ var tsCompileOptions = {
...
+    for (var i = 0; i < process.argv.length; i++) {
+        if (process.argv[i].indexOf(optionName) === 0) {
+            if (i + 1 < process.argv.length) {
+                return process.argv[i + 1];
+            }
+            break;
+        }
+    }
+    return defaultValue;
+}
```
geof90: Are the above two methods used at all?

scottbommarito: they were used by the emulator code, which is now in an entirely different file lol

test/testUtil.ts
```diff
@@ -9,45 +9,180 @@ import path = require("path");
...
-    public static PLATFORM_OPTION_NAME: string = "--platform";
-    public static TARGET_OPTION_NAME: string = "--target";
+    public static ANDROID_PLATFORM_OPTION_NAME: string = "--android";
+    public static ANDROID_SERVER_URL: string = "--androidserver";
+    public static ANDROID_EMULATOR: string = "--androidemu";
+    public static IOS_PLATFORM_OPTION_NAME: string = "--ios";
+    public static IOS_SERVER_URL: string = "--iosserver";
+    public static IOS_EMULATOR: string = "--iosemu";
+    public static RESTART_EMULATORS: string = "--clean";
     public static SHOULD_USE_WKWEBVIEW: string = "--use-wkwebview";
```
geof90: I see now that this flag is a tristate ("true", "both", "false"). Should we have another "--use-uiwebview", so that if both are specified, it will run the tests for both, and if none are specified, run the tests for uiwebview by default?

scottbommarito: I want the default behavior to be false. If I changed it to two flags, we could potentially have a situation where neither flags are specified, which wouldn't make much sense because we can't run tests without either webviews. Ultimately I like it better this way because the person has to know about the WkWebView to test for it.

geof90:
LGTM!

Issue #103, opened 2016-04-25
cordova-plugin-code-push@1.7.0-beta Cannot call "reportStatus" on undefined


chris-vaszauskas:
When I try to upload a build to Android, the update downloads successfully and running `codePush.restartApplication()` loads from the new code. However, when I force kill the app and open it again, I get a console error that says `Uncaught TypeError: Cannot read property 'reportStatus' of undefined` before any of my app code runs. I'm assuming this is the `reportStatus` function on the `codePush` object. My hunch is that the reportStatus function is getting called before the Cordova `deviceready` event when the codePush object becomes available.

When I downgrade to `cordova-plugin-code-push@1.6.0-beta`, I don't get the error.

I'm using the Crosswalk plugin at version 1.6.0.

lostintangent:
@chris-vaszauskas Interesting, we haven't seen this issue yet. Let me investigate this right now and get back to you. It's extremely helpful to know that downgrading to `v1.6.0` resolves it, and we'll make sure to try using CrossWalk in the app. Thanks for reporting the issue and sorry for the inconvenience.

lostintangent:
This issue has been fixed and we'll be releasing it to NPM later this evening. Thanks so much for submitting this!

chris-vaszauskas:
Thanks! That was fast!

PR #104, opened 2016-04-25
Fix timing issue with reporting status

ryuyu:
Fixes an issue with dependcy versions that could prevent an "npm install" from working correctly.
Fixes a reference case sensitivity issue in one of the test files.
Wraps reportStatus from native in a deviceready listener to prevent it from being called before codePush object is ready.

Should address issue https://github.com/Microsoft/cordova-plugin-code-push/issues/103

@lostintangent @dlebu @geof90

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": ">=3.0.0",
+    "cordova-plugin-file": "^3.0.0",
     "cordova-plugin-zip": ">=3.0.0",
     "cordova-plugin-dialogs": ">=1.1.1",
     "cordova-plugin-device": ">=1.1.0"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": "^3.0.0",
+    "cordova-plugin-file": ">=3.0.0",
     "cordova-plugin-zip": ">=3.0.0",
     "cordova-plugin-dialogs": ">=1.1.1",
     "cordova-plugin-device": ">=1.1.0"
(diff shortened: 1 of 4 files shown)


src/android/CodePushReportingManager.java
```diff
@@ -48,7 +48,7 @@ public void reportStatus(Status status, String label, String appVersion, String
         /* JS function to call: window.codePush.reportStatus(status: number, label: String, appVersion: String, currentDeploymentKey: String, previousLabelOrAppVersion?: string, previousDeploymentKey?: string) */
         final String script = String.format(
             Locale.US,
-            "javascript:window.codePush.reportStatus(%d, %s, %s, %s, %s, %s)",
+            "javascript:document.addEventListener(\"deviceready\", reportStatus); function reportStatus() { window.codePush.reportStatus(%d, %s, %s, %s, %s, %s); }",
```
lostintangent: Would it be cleaner to do `addEventListener("deviceready", codePush.reportStatus.bind(codePush, %d, %s, %s, %s, %s)`? That way you don't need the extra inline method definition?

geof90: +1

geof90:
LGTM!

package.json
```diff
@@ -24,7 +24,7 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": ">=3.0.0",
+    "cordova-plugin-file": "^3.0.0",
```
scottbommarito: if we're going to fix the cordova dependency in place to get around the file plugin bug, then I suggest fixing it to the commit that fixes the marshmallow permissions issue and verifying that it was before the bug that breaks the plugin. otherwise, my automated tests will not work.

scottbommarito: https://github.com/Microsoft/cordova-plugin-code-push/issues/57

scottbommarito: also, are you sure this even downloads version 3.0.0 of the plugin? my understanding was that it downloads the file-transfer plugin first because it's above, and since the file transfer plugin has a dependency on it, it downloads the most recent file plugin because that's how the file transfer plugin's dependencies are set. I would also set this to `^4.1.1` instead of `^3.0.0` because I believe that's the most recent version that wasn't broken.

lostintangent:
Can you replicate this change on iOS as well? The logic is almost identical, so it's a simple port.

package.json
lostintangent: I would vote for leaving this for now. As Scott mentioned, this fix is necessary for other reasons, so I'd rather change this dependency in another PR if we needed to.

lostintangent:
LGTM! If you can test this against iOS with both UI and WKWebView, and Android with both standard or CrossWalk, that would be awesome. We've had a handful of WebView regressions lately, so I want to make sure we don't introduce any others :)

dlebu:
LGTM!

Issue #108, opened 2016-04-28
Native crash on iOS when specifying deploymentKey without config.xml entry


Justin-Credible:
I'm using the `codePush.sync(...)` method with `SyncOptions` to pass my `deploymentKey`. I do not have a deployment key in my `config.xml` file.

This causes a crash on iOS the first time the app is launched. The crash occurs in `CodePushReportingManager.m` [line 33](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/src/ios/CodePushReportingManager.m#L33) when it attempts to insert a nil entry (the deployment key) into a dictionary:

````
 *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]'
````

Walking up the call stack to `-[CodePush updateSuccess:]` in `CodePush.m` we can see, on [line 64](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/src/ios/CodePush.m#L64), that the deployment key is pulled from `config.xml` directly (which is null in my case). I would expect this to use the value I passed into the sync method.

This is similar to https://github.com/Microsoft/cordova-plugin-code-push/issues/89, but instead of errors in the console, a native code crash occurs. I can use the workaround listed there to bypass the crash, but the deployment key would still be incorrect (it would be empty string instead of nil).

Version: v1.7.0-beta

Steps to reproduce:
1. Ensure the app is NOT installed on the phone
2. Make sure the deployment key is not in `config.xml`
3. Add a call to `codePush.sync(...)` that passed `SyncOptions` with a `deploymentKey` set
4. Launch the app

Expected Results: No crash and the `updateSuccess` method uses the deployment key I specified.

Actual Results: Native code crash because the deployment key is not present in `config.xml`

chris-vaszauskas:
I'm seeing this also.

lostintangent:
Hey guys! We can repro this bug, and as you mentioned, the workaround in the meantime is to simply add a deployment key to your `config.xml`. The key that you specify to the call to `sync` will be the one that is used, so everything should work as expected. It's a non-ideal solution, but it should be impactless on your app. Let us know if that unblocks you.

In parallel, we'll investigate a fix for this issue and try to get it out ASAP! Thanks for reporting this.

Justin-Credible:
[...]

PR #109, opened 2016-04-28
Fix nil deploymentkey crash

ryuyu:
Fixes a crash that was caused by attempting to report status when the deployment key wasn't specified in config.xml.

Addresses https://github.com/Microsoft/cordova-plugin-code-push/issues/108

@lostintangent @geof90

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -63,7 +63,7 @@ public class CodePush extends CordovaPlugin {
         } else if ("install".equals(action)) {
             return execInstall(args, callbackContext);
         } else if ("updateSuccess".equals(action)) {
-            return execUpdateSuccess(callbackContext);
+            return execUpdateSuccess(args, callbackContext);
         } else if ("restartApplication".equals(action)) {
             return execRestartApplication(args, callbackContext);
         } else if ("isPendingUpdate".equals(action)) {
@@ -103,13 +103,18 @@ public class CodePush extends CordovaPlugin {
         return true;
     }
 
-    private boolean execUpdateSuccess(CallbackContext callbackContext) {
+    private boolean execUpdateSuccess(CordovaArgs args, CallbackContext callbackContext) {
+        String deploymentKey = args.getString(0);
+        if (deploymentKey.equals(null) || string.equals("") || string.equals("null")) {
+            deploymentKey = mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null);
+        }
+
         if (this.codePushPackageManager.isFirstRun()) {
             this.codePushPackageManager.saveFirstRunFlag();
             /* save reporting status for first install */
             try {
                 String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
+                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, deploymentKey, this.mainWebView);
             } catch (PackageManager.NameNotFoundException e) {
                 // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
                 e.printStackTrace();
@@ -310,13 +315,6 @@ public class CodePush extends CordovaPlugin {
                             this.codePushPackageManager.clearFailedUpdates();
                             this.codePushPackageManager.clearPendingInstall();
                             this.codePushPackageManager.clearInstallNeedsConfirmation();
-                            try {
-                                String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
-                            } catch (PackageManager.NameNotFoundException e) {
-                                // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
-                                e.printStackTrace();
-                            }
                         }
                     }
                 }
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -104,8 +104,14 @@ public class CodePush extends CordovaPlugin {
     }
 
     private boolean execUpdateSuccess(CordovaArgs args, CallbackContext callbackContext) {
-        String deploymentKey = args.getString(0);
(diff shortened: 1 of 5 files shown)


src/ios/CodePush.m
```diff
@@ -58,10 +58,14 @@ - (void)handleUnconfirmedInstall:(BOOL)navigate {
 }
 
 - (void)updateSuccess:(CDVInvokedUrlCommand *)command {
+    NSString* deploymentKey = [command argumentAtIndex:0 withDefault:nil andClass:[NSString class]];
+    if ([deploymentKey length] == 0) {
```
dlebu: Is this `nil` proof?

typings/codePush.d.ts
```diff
@@ -187,10 +187,11 @@ interface CodePushCordovaPlugin {
      * Calling this function is required on the first run after an update. On every subsequent application run, calling this function is a noop.
      * If using sync API, calling this function is not required since sync calls it internally.
      *
+     * @param deploymentKey Optional string deployment key for this package.
      * @param notifySucceeded Optional callback invoked if the plugin was successfully notified.
      * @param notifyFailed Optional callback invoked in case of an error during notifying the plugin.
      */
-    notifyApplicationReady(notifySucceeded?: SuccessCallback<void>, notifyFailed?: ErrorCallback): void;
+    notifyApplicationReady(deploymentKey?: string, notifySucceeded?: SuccessCallback<void>, notifyFailed?: ErrorCallback): void;
```
dlebu: This is an API breaking change. Is this really necessary? Can't we be smarter in the native code and remember the custom deployment key?

www/codePush.ts
```diff
@@ -43,11 +43,12 @@ class CodePush implements CodePushCordovaPlugin {
      * Calling this function is required on the first run after an update. On every subsequent application run, calling this function is a noop.
      * If using sync API, calling this function is not required since sync calls it internally.
      *
+     * @param deploymentKey Optional string deployment key for this package.
      * @param notifySucceeded Optional callback invoked if the plugin was successfully notified.
      * @param notifyFailed Optional callback invoked in case of an error during notifying the plugin.
      */
-    public notifyApplicationReady(notifySucceeded?: SuccessCallback<void>, notifyFailed?: ErrorCallback): void {
-        cordova.exec(notifySucceeded, notifyFailed, "CodePush", "updateSuccess", []);
+    public notifyApplicationReady(deploymentKey?: string, notifySucceeded?: SuccessCallback<void>, notifyFailed?: ErrorCallback): void {
```
dlebu: JS source file should be checked in as well.

typings/codePush.d.ts
scottbommarito: I don't mind adding deploymentKey to notifyApplicationReady, but it needs to be the last argument. Not only does it break the API if it's at the front, but it's more consistent with checkForUpdate's custom deployment key specification if it's at the back.

src/ios/CodePush.m
ryuyu: Yes. Calling length on nil will return 0.

typings/codePush.d.ts
ryuyu: Yes I forgot to consider that. I will just move it to the end.

dlebu: I think, as much as possible, we should keep our interfaces for react and cordova in sync. AFAIK, the react plugin does not expose this parameter. @scottbommarito We should not do it if it is not needed.

src/android/CodePush.java
```diff
@@ -103,13 +103,24 @@ protected Void doInBackground(Void... params) {
...
-    private boolean execUpdateSuccess(CallbackContext callbackContext) {
+    private boolean execUpdateSuccess(CordovaArgs args, CallbackContext callbackContext) {
+        String deploymentKey = "";
+        try {
+            deploymentKey = args.getString(0);
+        } catch (JSONException e) {
+            // no-op becuase we don't care if we didn't get a deployment key from script; we will just try to use the one from preferences
+        }
+
+        if (deploymentKey.equals(null) || deploymentKey.equals("") || deploymentKey.equals("null") || deploymentKey.equals("undefined")) {
```
geof90: `deploymentKey.equals(null)` will never be true. If `deploymentKey == null`,  `deploymentKey.equals(null)` will throw a `NullPointerException`.

ryuyu: Right. Forogt about that. Fixed.

scottbommarito:
imo I think this problem is a little more complex than this solution.

basically, we now have 3 separate deployment keys
1 - deployment key in config.xml
2 - deployment key passed to checkForUpdate
3 - "default" deployment key (used when 1 and 2 are unspecified)

right now, the default key is null. when we don't have a key in config.xml and no key was passed into checkForUpdate, then we crash.

I agree with @dlebu 's assessment that we can be smarter in our native code to prevent the default key from being null. I like adding the deployment key as an argument to notifyApplicationReady, but I think that if the deployment key is unspecified, it should use the key used in the last call to checkForUpdate (when you call notifyApplicationReady you're probably trying to validate the last update u checked for). If not, it should use the key in the config.xml.

src/android/CodePush.java
```diff
@@ -310,13 +321,6 @@ private void handleAppStart() {
                             this.codePushPackageManager.clearFailedUpdates();
                             this.codePushPackageManager.clearPendingInstall();
                             this.codePushPackageManager.clearInstallNeedsConfirmation();
-                            try {
-                                String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
-                            } catch (PackageManager.NameNotFoundException e) {
-                                // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
-                                e.printStackTrace();
-                            }
```
geof90: With this code removed, wouldn't that mean we wont be able to report binary updates?

www/codePush.ts
```diff
@@ -293,7 +294,7 @@ class CodePush implements CodePushCordovaPlugin {
             CodePushUtil.copyUnassignedMembers(defaultOptions, syncOptions);
         }
 
-        window.codePush.notifyApplicationReady();
+        window.codePush.notifyApplicationReady(null, null, syncOptions.deploymentKey);
```
geof90: Can't we make the deploymentKey the first param? Then we wouldn't have to do `null, null`. Seems more idiomatic for node style async callback functions too.

geof90: Sorry, just saw that @dlebu commented on this previously.

ryuyu:
Pulling this PR in favor of minimum change required to unblock.

PR #110, opened 2016-04-28
Work around crash with no deployment key in config.xml

ryuyu:
The crash itself was coming from attempting to save away some information in an NSDictionary with a nil value.

After some discussion, we decided that at this time, it would be best to no-op the reporting in the case where config.xml deployment key is not set, as this is consistent with the behaviour of the react native plugin, and it unblocks this scenario.

@lostintangent @dlebu @geof90


geof90:
LGTM!

lostintangent:
LGTM!

dlebu:
LGTM!

shishirx34:
LGTM

PR #115, opened 2016-05-04
Added test reporter so that tests can report their success/failure rate

scottbommarito:
this is coupled with some changes on TFS that make the build output much friendlier

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,15 @@ Follow these steps to build a new version of the plugin:
 
 ### Environment setup
 
-Make sure you can build the plugin by following the steps above.
+First, make sure you can build the plugin by following the steps above.
+
+Then, make sure you have installed ```Gulp```, ```Mocha```, and ```Mocha JUnit Reporter``` through npm.
+
+```
+npm install -g gulp
+npm install -g mocha
+npm install -g mocha-junit-reporter
+```
 
 To run Android tests, make sure you have ```sdk\tools``` and  ```sdk\platform-tools``` in your PATH.
 
@@ -67,6 +75,8 @@ If you would like to skip building, add a ```-fast``` to the end of the command 
 There is a both a full unit test suite and a "core" set of unit tests that you may run. If you would like to run only the core tests, add a ```--core``` flag to the command.
 
 If you would like to pull the plugin from NPM rather than running the tests on the local version, add a ```--npm``` flag to the command.
+
+The mocha reporter outputs individual results files for each platform. These are ```./test_android.xml```, ```./test-ios-ui.xml```, and ```./test-ios-wk.xml```.
 
 #### Default
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,13 @@ Instead, modify the sources in the ```www``` folder and compile a new version of
 
 ### Environment setup
 
-```node.js``` and ```npm``` are needed for building this project. ```npm``` comes bundled with the ```node.js``` installer. You can download the ```node.js``` installer here: https://nodejs.org/download/. 
+```node.js``` and ```npm``` are needed for building this project. ```npm``` comes bundled with the ```node.js``` installer. You can download the ```node.js``` installer here: https://nodejs.org/download/.
+
+Once you have installed ```node.js``` and ```npm```, install the dev dependencies for the project.
+
+```
+npm install
+```
 
 ### Compile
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -33,7 +33,15 @@ Follow these steps to build a new version of the plugin:
...
 
-Make sure you can build the plugin by following the steps above.
+First, make sure you can build the plugin by following the steps above.
+
+Then, make sure you have installed ```Gulp```, ```Mocha```, and ```Mocha JUnit Reporter``` through npm.
+
+```
+npm install -g gulp
+npm install -g mocha
+npm install -g mocha-junit-reporter
```
dlebu: Can we add these to `devDependencies` and run them from the `bin` location? Gulp is probably required to be global, but the others most likely don't. For example, for a local `mocha` install, you can run it from `node_modules/mocha/bin/mocha` without having to install it globally. This way, developers do `npm install` and they are done with the setup, since that installs devDependencies directly.

dlebu: Another location for the mocha example is `node_modules/.bin/mocha`. This contains all the executables for the installed modules.

scottbommarito: hmm gulp is already in devdependencies but mocha is not...I recall having to install gulp and mocha globally to get it to work in the first place, i'll investigate how changing the devdependencies affects it

scottbommarito: so it works with node_modules/.bin/mocha but you would have to call node_modules/.bin/gulp if you want that to work off devdependencies alone, so I will keep the gulp install in there and then tell the user to do an npm install

package.json
```diff
@@ -24,7 +24,6 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": ">=3.0.0",
```
scottbommarito: I changed this because `npm install` was giving errors constantly about the installed version of `cordova-plugin-file` not satisfying `cordova-plugin-file-transfer`. Is this the right way to do this?

scottbommarito:
@dlebu @lostintangent @geof90 
someone pls review lol

package.json
geof90: Does this problem still show up? I noticed that the file transfer plugin bumped their dependency version of the file plugin to ^4.0.0. https://github.com/apache/cordova-plugin-file-transfer/commit/3540af439ed0a91885ff2cbf5e0a3d42c7fd05ed#diff-53f390d375398624afe1cfe1125f42bfR32

As part of a seperate discussion, I think we should change all of our usage of `>=` to `^`.

gulpfile.js
```diff
@@ -111,10 +111,19 @@ function execCommandWithPromise(command, options, logOutput) {
...
     
-    // pass arguments supplied by test tasks
+    // Set up the mocha junit reporter.
+    args.push("--reporter");
+    args.push("mocha-junit-reporter");
+    // Set the mocha reporter to the correct output file.
+    args.push("--reporter-options");
+    if (options.android && !options.ios) args.push("mochaFile=./test-android.xml");
+    else if (options.ios && !options.android) args.push("mochaFile=./test-ios" + (options.wk ? (options.ui ? "" : "-wk") : "-ui") + ".xml");
+    else args.push("mochaFile=./test-results.xml");
```
geof90: nit: For long lines like these three above, I prefer if you used braces.

geof90:
LGTM!

CONTRIBUTING.md
```diff
@@ -33,7 +39,13 @@ Follow these steps to build a new version of the plugin:
 
 ### Environment setup
 
-Make sure you can build the plugin by following the steps above.
+First, make sure you can build the plugin by following the steps above.
+
+Then, make sure you have installed ```Gulp```.
```
dlebu: nit: you only need one set of ticks for inline formatting `Gulp` = \`Gulp\`

package.json
dlebu: I agree, we should not make this change of part of this PR.

dlebu:
I had one comment about the `package.json` peer dependency change, which should not be part of this PR, as @geof90 said. Other than that, LGTM!

CONTRIBUTING.md
scottbommarito: LOL I saw the triple ticks for some inline code snippet somewhere and started using it everywhere...time to fix!

PR #119, opened 2016-05-11
fix WkWebView tests

scottbommarito:
for some reason, the wkwebview tests are sending requests with X-CodePush-SDK-Version headers while the uiwebview tests do not, which the server was not configured to handle

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -169,7 +169,7 @@ function runTests(targetPlatform: platform.IPlatform, useWkWebView: boolean): vo
         app.use(function(req: any, res: any, next: any) {
             res.setHeader("Access-Control-Allow-Origin", "*");
             res.setHeader("Access-Control-Allow-Methods", "*");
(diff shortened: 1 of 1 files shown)


scottbommarito:
@dlebu @geof90

dlebu:
LGTM!

geof90:
Are you sure the UIWebView isn't sending it? It should/needs send it, and it would be a bug if it didn't. https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/httpRequester.ts#L43

scottbommarito:
@geof90 
if it didn't, where would we see errors?

geof90:
@scottbommarito We won't see errors, but the CodePush server requires it for the metrics feature.
Can you double check that it is indeed sending it?

scottbommarito:
@geof90 
just checked, it does send!

geof90:
LGTM!

PR #120, opened 2016-05-11
fix small bugs in testing

scottbommarito:
file transfer includes file@^3.0.0 as a dependency, but we install file@>=3.0.0. As a result, file transfer is incompatible with the version of file that we install and makes npm install exit with nonzero, even though it still installs everything and you can run everything fine. This fix keeps our dependency on file while allowing us to not have to update it to keep it consistent with the version that file-transfer depends on

Additionally, I removed the RESTART_SUCCEEDED message from the restartApplication test case. It was not being reliably sent on wkwebview (likely due to the webview being killed before it sends completely since it performs faster than uiwebview), and since when the restart step fails it will still call RESTART_FAILED and fail the test it's not necessary.

with this merged, the automated tests finally completely succeed!
https://mobiledevops.visualstudio.com/VS%20Client/_build?_a=summary&buildId=3908
https://mobiledevops.visualstudio.com/_permalink/_build/index?collectionId=ffea02da-5599-45aa-a688-8709d085b2fb&projectId=5fc0653c-f92f-4d49-9e5e-a6aa08350131&buildId=3909

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,6 @@
     "cordova-android",
     "cordova-ios"
   ],
-  "peerDependencies": {
-    "code-push": ">=1.8.0-beta",
-    "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-zip": ">=3.0.0",
-    "cordova-plugin-dialogs": ">=1.1.1",
-    "cordova-plugin-device": ">=1.1.0"
-  },
   "author": "Microsoft Corporation",
   "license": "MIT",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": ">=3.0.0",
     "cordova-plugin-zip": ">=3.0.0",
     "cordova-plugin-dialogs": ">=1.1.1",
     "cordova-plugin-device": ">=1.1.0"
--- a/test/template/www/js/scenarioRestart.js
+++ b/test/template/www/js/scenarioRestart.js
@@ -36,7 +36,8 @@ var app = {
         window.codePush.restartApplication(
             function () {
                 /* success */
-                app.sendTestMessage("RESTART_SUCCEEDED");
+                // does not always send...rely on RESTART_FAILED to provide failure state instead of checking for RESTART_SUCCEEDED
+                // app.sendTestMessage("RESTART_SUCCEEDED");
                 callback && callback();
             },
             function () {
--- a/test/template/www/js/scenarioRestart.js
+++ b/test/template/www/js/scenarioRestart.js
@@ -35,9 +35,6 @@ var app = {
     tryRestart: function (callback) {
         window.codePush.restartApplication(
             function () {
-                /* success */
-                // does not always send...rely on RESTART_FAILED to provide failure state instead of checking for RESTART_SUCCEEDED
-                // app.sendTestMessage("RESTART_SUCCEEDED");
                 callback && callback();
             },
             function () {
--- a/test/test.ts
+++ b/test/test.ts
@@ -893,7 +893,7 @@ function runTests(targetPlatform: platform.IPlatform, useWkWebView: boolean): vo
                             new su.AppMessage(su.TestMessage.SYNC_STATUS, [su.TestMessage.SYNC_UPDATE_INSTALLED]),
                             new su.AppMessage(su.TestMessage.PENDING_PACKAGE, [mockResponse.updateInfo.packageHash]),
                             new su.AppMessage(su.TestMessage.CURRENT_PACKAGE, [null]),
-                            su.TestMessage.RESTART_SUCCEEDED,
+                            // su.TestMessage.RESTART_SUCCEEDED,
                             su.TestMessage.DEVICE_READY_AFTER_UPDATE,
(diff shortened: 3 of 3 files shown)


scottbommarito:
@geof90 @dlebu

scottbommarito:
@geof90 @dlebu 
pinging again because I want to get this merged asap

package.json
```diff
@@ -24,7 +24,6 @@
   "peerDependencies": {
     "code-push": ">=1.8.0-beta",
     "cordova-plugin-file-transfer": ">=1.3.0",
-    "cordova-plugin-file": ">=3.0.0",
```
dlebu: Cordova actually looks [here](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/plugin.xml#L10) when installing dependencies, so the fix for this would be to adjust the version numbers (in both `package.json` and `plugin.xml` to stay in sync, and keeping in mind what the file transfer plugin depends on) and not to remove the peer dependency. Would that work?

dlebu: The latest file transfer plugin depends on file `^4.0.0`. https://github.com/apache/cordova-plugin-file-transfer/blob/master/plugin.xml#L32 Do we want to upgrade to that? What was the version of the file plugin that fixed our Android permissions issue?

scottbommarito: Hey Daniel, we looked into this and it looks like the reason why this is broken is that their [package.json](https://github.com/apache/cordova-plugin-file-transfer/blob/1.5.x/package.json#L48) does not match their plugin.xml, leading to npm giving an error. They appear to have fixed it on their current dev branch but it's not on npm.

Do you even need to have the cordova plugins in package.json as peer dependencies in order to build? Aren't they only needed when cordova installs the plugin?

geof90: What's the drawback of just removing all the peerDependencies? Are they even needed?

scottbommarito: Just checked--it builds and runs without any peer dependencies. Additionally, the fix that [cordova-plugin-file-transfer](https://github.com/apache/cordova-plugin-file-transfer/blob/master/package.json) made was to remove all of their peer dependencies.

geof90: @dlebu Do u know if VS is using the `peerDependencies` in `package.json`? Otherwise I think the right fix is to just get rid of all the `peerDependencies` since `plugin.xml` should be sufficient for Cordova.

dlebu: I don't think they do, but it's worth double checking. I am fine with removing them completely - cordova should still be able to work without them. We might want to consider bumping the file plugin version in `config.xml` to take version of the plugin that contains the Android permissions fix.

scottbommarito: How would we check if they do? I'm browsing through the docs and I don't see anything

dlebu: You could try it out in the product and see if the dependencies are installed. Given that a core plugin (file transfer) did the same thing, I think we're okay.

scottbommarito: alright, sounds good

test/template/www/js/scenarioRestart.js
```diff
@@ -36,7 +36,8 @@ var app = {
         window.codePush.restartApplication(
             function () {
                 /* success */
-                app.sendTestMessage("RESTART_SUCCEEDED");
+                // does not always send...rely on RESTART_FAILED to provide failure state instead of checking for RESTART_SUCCEEDED
+                // app.sendTestMessage("RESTART_SUCCEEDED");
```
geof90: I think it doesn't send because sendTestMessage is async but you are then subsequently invoking the callback synchronously. Can we add a callback to `sendTestMessage` and then call the below `callback && callback()` after the message is sent?

scottbommarito: good point, I will check this. I wonder why it only breaks with wkwebview then

scottbommarito: actually, just kidding, the callback has nothing to do with it, but making the send synchronous may fix it

scottbommarito: just kidding, the send is already synchronous (note the `false` as the third parameter which denotes a synchronous request) so this sounds like a bug in the wkwebview to me

scottbommarito:
@geof90 @dlebu 
I tested the plugin without peer dependencies through VS and it worked. Should be good to merge!

test/template/www/js/scenarioRestart.js
geof90: Can you check that its working? It might be removed in WKWebView. https://xhr.spec.whatwg.org/#synchronous-flag

> Synchronous XMLHttpRequest outside of workers is in the process of being removed from the web platform as it has detrimental effects to the end user's experience. (This is a long process that takes many years.) Developers must not pass false for the async argument when entry settings object's global object is a Window object. User agents are strongly encouraged to warn about such usage in developer tools and may experiment with throwing an InvalidAccessError exception when it occurs.

geof90: But I'm ok with merging this change if it's easier. I guess its not super important. I'll leave it up to you.

geof90:
LGTM!

test/template/www/js/scenarioRestart.js
scottbommarito: Interesting...the thing is, if that error was the problem, it would be throwing in every test case, since they're all synchronous. I think this is just an issue with wkwebview. I'm just going to merge it as-is--I've seen all of the platforms not send lifecycle test messages (APPLICATION_RESUMED) at some point, which is why I've removed all of them from the tests and this is just the last one to go lol

PR #121, opened 2016-05-13
Retry reportStatus if failure

geof90:
This PR improves the reportStatus mechanism in order to make sure that metrics reporting calls are not lost due to bad network conditions, hence prevents metrics numbers from being "corrupted".

In a nutshell, if the call to the server fails/times out, we will save a copy of the statusReport in the Preferences, and retry again the next time the app resumes.

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -46,32 +46,36 @@ public class CodePush extends CordovaPlugin {
 
     @Override
     public boolean execute(String action, CordovaArgs args, final CallbackContext callbackContext) {
-        if ("getServerURL".equals(action)) {
-            this.returnStringPreference("codepushserverurl", callbackContext);
-            return true;
+        if ("getAppVersion".equals(action)) {
+            return execGetAppVersion(callbackContext);
+        } else if ("getBinaryHash".equals(action)) {
+            return execGetBinaryHash(callbackContext);
         } else if ("getDeploymentKey".equals(action)) {
             this.returnStringPreference(DEPLOYMENT_KEY_PREFERENCE, callbackContext);
             return true;
         } else if ("getNativeBuildTime".equals(action)) {
             return execGetNativeBuildTime(callbackContext);
-        } else if ("getAppVersion".equals(action)) {
-            return execGetAppVersion(callbackContext);
-        } else if ("getBinaryHash".equals(action)) {
-            return execGetBinaryHash(callbackContext);
-        } else if ("preInstall".equals(action)) {
-            return execPreInstall(args, callbackContext);
+        } else if ("getServerURL".equals(action)) {
+            this.returnStringPreference("codepushserverurl", callbackContext);
+            return true;
         } else if ("install".equals(action)) {
             return execInstall(args, callbackContext);
-        } else if ("updateSuccess".equals(action)) {
-            return execUpdateSuccess(callbackContext);
-        } else if ("restartApplication".equals(action)) {
-            return execRestartApplication(args, callbackContext);
-        } else if ("isPendingUpdate".equals(action)) {
-            return execIsPendingUpdate(args, callbackContext);
         } else if ("isFailedUpdate".equals(action)) {
             return execIsFailedUpdate(args, callbackContext);
         } else if ("isFirstRun".equals(action)) {
             return execIsFirstRun(args, callbackContext);
+        } else if ("isPendingUpdate".equals(action)) {
+            return execIsPendingUpdate(args, callbackContext);
+        } else if ("preInstall".equals(action)) {
+            return execPreInstall(args, callbackContext);
+        } else if ("reportFailed".equals(action)) {
+            return execReportFailed(args, callbackContext);
+        } else if ("reportSucceeded".equals(action)) {
+            return execReportSucceeded(args, callbackContext);
+        } else if ("restartApplication".equals(action)) {
+            return execRestartApplication(args, callbackContext);
+        } else if ("updateSuccess".equals(action)) {
+            return execUpdateSuccess(callbackContext);
         } else {
             return false;
         }
@@ -109,17 +113,17 @@ public class CodePush extends CordovaPlugin {
             /* save reporting status for first install */
             try {
                 String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
+                codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null)), this.mainWebView);
             } catch (PackageManager.NameNotFoundException e) {
                 // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
                 e.printStackTrace();
             }
-        }
-
-        if (this.codePushPackageManager.installNeedsConfirmation()) {
+        } else if (this.codePushPackageManager.installNeedsConfirmation()) {
             /* save reporting status */
             CodePushPackageMetadata currentMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
-            codePushReportingManager.reportStatus(CodePushReportingManager.Status.UPDATE_CONFIRMED, currentMetadata.label, currentMetadata.appVersion, currentMetadata.deploymentKey, this.mainWebView);
+            codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.UPDATE_CONFIRMED, currentMetadata.label, currentMetadata.appVersion, currentMetadata.deploymentKey), this.mainWebView);
+        } else if (codePushReportingManager.hasFailedReport()) {
+            codePushReportingManager.reportStatus(codePushReportingManager.getAndClearFailedReport(), this.mainWebView);
         }
 
         this.codePushPackageManager.clearInstallNeedsConfirmation();
@@ -194,6 +198,28 @@ public class CodePush extends CordovaPlugin {
         } catch (Exception e) {
             callbackContext.error("Cound not read webview URL: " + e.getMessage());
         }
+        return true;
+    }
+
+    private boolean execReportFailed(CordovaArgs args, CallbackContext callbackContext) {
+        try {
+            StatusReport statusReport = StatusReport.deserialize(args.optJSONObject(0));
+            codePushReportingManager.saveFailedReport(statusReport);
+        } catch (JSONException e) {
+            Utilities.logException(e);
+        }
+
+        return true;
+    }
+
+    private boolean execReportSucceeded(CordovaArgs args, CallbackContext callbackContext) {
+        try {
+            StatusReport statusReport = StatusReport.deserialize(args.optJSONObject(0));
+            codePushReportingManager.saveSuccessfulReport(statusReport);
+        } catch (JSONException e) {
+            Utilities.logException(e);
+        }
+
         return true;
     }
 
@@ -312,7 +338,7 @@ public class CodePush extends CordovaPlugin {
                             this.codePushPackageManager.clearInstallNeedsConfirmation();
                             try {
                                 String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
+                                codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null)), this.mainWebView);
                             } catch (PackageManager.NameNotFoundException e) {
                                 // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
                                 e.printStackTrace();
@@ -330,7 +356,7 @@ public class CodePush extends CordovaPlugin {
         if (this.codePushPackageManager.installNeedsConfirmation()) {
             /* save status for later reporting */
             CodePushPackageMetadata currentMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
-            codePushReportingManager.reportStatus(CodePushReportingManager.Status.UPDATE_ROLLED_BACK, currentMetadata.label, currentMetadata.appVersion, currentMetadata.deploymentKey, this.mainWebView);
+            codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.UPDATE_ROLLED_BACK, currentMetadata.label, currentMetadata.appVersion, currentMetadata.deploymentKey), this.mainWebView);
 
             /* revert application to the previous version */
             this.codePushPackageManager.clearInstallNeedsConfirmation();
@@ -461,6 +487,8 @@ public class CodePush extends CordovaPlugin {
                 handleAppStart();
                 this.markUpdate();
                 this.codePushPackageManager.clearPendingInstall();
+            } else if (codePushReportingManager.hasFailedReport()) {
+                codePushReportingManager.reportStatus(codePushReportingManager.getAndClearFailedReport(), this.mainWebView);
             }
         }
     }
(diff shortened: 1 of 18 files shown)


src/android/CodePush.java
```diff
@@ -46,32 +46,36 @@ public void initialize(CordovaInterface cordova, CordovaWebView webView) {
 
     @Override
     public boolean execute(String action, CordovaArgs args, final CallbackContext callbackContext) {
-        if ("getServerURL".equals(action)) {
-            this.returnStringPreference("codepushserverurl", callbackContext);
-            return true;
+        if ("getAppVersion".equals(action)) {
```
scottbommarito: actually, can we change this awful if/else chain into a switch statement? thanks for alphabetizing btw

src/android/CodePush.java
```diff
@@ -109,17 +113,17 @@ private boolean execUpdateSuccess(CallbackContext callbackContext) {
...
-                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
+                codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null)), this.mainWebView);
             } catch (PackageManager.NameNotFoundException e) {
                 // Should not happen unless the appVersion is not specified, in which case we can't report anything anyway.
                 e.printStackTrace();
             }
-        }
-
-        if (this.codePushPackageManager.installNeedsConfirmation()) {
+        } else if (this.codePushPackageManager.installNeedsConfirmation()) {
```
scottbommarito: this appears to be a big logic change, from if to else if, what is the context behind this that it doesn't break the previous behavior?

src/android/CodePush.java
```diff
@@ -109,17 +113,17 @@ private boolean execUpdateSuccess(CallbackContext callbackContext) {
             /* save reporting status for first install */
             try {
                 String appVersion = Utilities.getAppVersionName(cordova.getActivity());
-                codePushReportingManager.reportStatus(CodePushReportingManager.Status.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null), this.mainWebView);
+                codePushReportingManager.reportStatus(new StatusReport(ReportingStatus.STORE_VERSION, null, appVersion, mainWebView.getPreferences().getString(DEPLOYMENT_KEY_PREFERENCE, null)), this.mainWebView);
```
scottbommarito: i think this method should be renamed from `execUpdateSuccess` to something more telling of what it actually does. as far as i can tell, it's more of a `execConfirmInstall`. the issue is that  `execUpdateSuccess` implies that an update has just been installed and succeeded, but that is not always the case--if i'm correct, it's called every time notifyApplicationReady happens, which means any number of situations could be the case, most of which are actually not the result of installing updates

additionally i think there needs to be a comment in here to explain this if/else chain behavior on a more general level
```
// Report the current installation to metrics.
// If it is the first run of a store version app, an installation that needs confirmation, or an installation for which the previous report failed, we need to report to metrics.
// If it is not any of those cases, the installation has already been reported to metrics so we don't need to do so again.
```

src/android/CodePush.java
```diff
@@ -461,6 +487,8 @@ public void onStart() {
                 handleAppStart();
                 this.markUpdate();
                 this.codePushPackageManager.clearPendingInstall();
+            } else if (codePushReportingManager.hasFailedReport()) {
```
scottbommarito: why can we only report a failed update on resume? why not on restart? are we waiting for the installation to be confirmed?

src/android/CodePushReportingManager.java
```diff
@@ -44,26 +22,22 @@ public CodePushReportingManager(Activity cordovaActivity, CodePushPreferences co
     /**
      * Invokes the window.codePush.reportStatus JS function for the given webView.
      */
-    public void reportStatus(Status status, String label, String appVersion, String deploymentKey, final CordovaWebView webView) {
+    public void reportStatus(StatusReport statusReport, final CordovaWebView webView) {
         /* JS function to call: window.codePush.reportStatus(status: number, label: String, appVersion: String, currentDeploymentKey: String, previousLabelOrAppVersion?: string, previousDeploymentKey?: string) */
-        if (deploymentKey == null || deploymentKey.isEmpty()) {
+        if (statusReport.deploymentKey == null || statusReport.deploymentKey.isEmpty()) {
```
scottbommarito: oh this reminds me of Yahoo...we had a Util class full of a bunch of wrapper methods for situations like this...
```
public static void isEmptyOrNull(string a) {
    return a == null || a.isEmpty();
}
```
not sure how many times we do checks like this but it may be worth looking into adding some utility functions

src/android/CodePushReportingManager.java
```diff
@@ -44,26 +22,22 @@ public CodePushReportingManager(Activity cordovaActivity, CodePushPreferences co
...
-            );
-
-        if (status == Status.STORE_VERSION || status == Status.UPDATE_CONFIRMED) {
-            codePushPreferences.saveLastVersion(label == null ? appVersion : label, deploymentKey);
-        }
+            statusReport.status.getValue(),
+            convertStringParameter(statusReport.label),
+            convertStringParameter(statusReport.appVersion),
+            convertStringParameter(statusReport.deploymentKey),
+            statusReport.lastVersionLabelOrAppVersion == null ? convertStringParameter(codePushPreferences.getLastVersionLabelOrAppVersion()) : convertStringParameter(statusReport.lastVersionLabelOrAppVersion),
```
scottbommarito: why would this be null, and why would it be correct to report if the version stored in codePushPreferences if it is?

both this and the line scare me in terms of an "we could get a really cryptic error due to reporting the wrong version or reporting when we shouldn't and i never want to deal with something like that"-kind of way

src/android/CodePushReportingManager.java
```diff
@@ -72,6 +46,34 @@ public void run() {
         });
     }
 
+    public boolean hasFailedReport() {
+        if (this.hasFailedReport == null) {
+            this.hasFailedReport = codePushPreferences.getFailedReport() != null;
+        }
+
+        return this.hasFailedReport;
```
scottbommarito: cool one-liner that's probably uglier than what u have but w/e:
`return this.hasFailedReport = this.hasFailedReport == null ? codePushPreferences.getFailedReport() != null : this.hasFailedReport;`

src/android/CodePushReportingManager.java
```diff
@@ -13,28 +12,7 @@
...
-        }
-
-        /**
-         * Returns the value associated with this enum.
-         */
-        public int getValue() {
-            return this.value;
-        }
-    }
+    private Boolean hasFailedReport = null;
```
scottbommarito: it seems like the only use of this is to avoid checking the preferences...but given that you don't report updates very often and this variable could get out of sync fairly easily i think it'd be better to remove this and just use codePushPreferences

src/android/StatusReport.java
```diff
@@ -0,0 +1,80 @@
+package com.microsoft.cordova;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class StatusReport {
```
scottbommarito: put a general "what this class does" comment
```
/**
 * Represents a status report to be sent to metrics.
 */
```
(the important part is that it specifies that it's for metrics, because otherwise there's absolutely context on what these reports are for)

src/android/StatusReport.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final String LAST_VERSION_DEPLOYMENT_KEY_KEY = "lastVersionDeploymentKey";
+
+    ReportingStatus status;
+    String label;
+    String appVersion;
+    String deploymentKey;
+    String lastVersionLabelOrAppVersion;
+    String lastVersionDeploymentKey;
+
+    public StatusReport(int status, String label, String appVersion, String deploymentKey, String lastVersionLabelOrAppVersion, String lastVersionDeploymentKey) {
```
scottbommarito: you may want a `public StatusReport(int status, String label, String appVersion, String deploymentKey)` as well

src/android/StatusReport.java
```diff
@@ -0,0 +1,80 @@
...
+    public StatusReport(ReportingStatus status, String label, String appVersion, String deploymentKey) {
+        this(status, label, appVersion, deploymentKey, null, null);
+    }
+
+    public StatusReport(ReportingStatus status, String label, String appVersion, String deploymentKey, String lastVersionLabelOrAppVersion, String lastVersionDeploymentKey) {
+        this.status = status;
+        this.label = label;
+        this.appVersion = appVersion;
+        this.deploymentKey = deploymentKey;
+        this.lastVersionLabelOrAppVersion = lastVersionLabelOrAppVersion;
```
scottbommarito: could we check if `lastVersionLabelOrAppVersion` and `lastVersionDeploymentKey` are null and fill them with the values from `codePushPreferences` instead of just having a null references here?

src/ios/CodePush.m
```diff
@@ -60,15 +62,17 @@ - (void)handleUnconfirmedInstall:(BOOL)navigate {
 - (void)updateSuccess:(CDVInvokedUrlCommand *)command {
```
scottbommarito: same comment as in android--tl;dr: name should be different

src/ios/CodePushReportingManager.m
```diff
@@ -1,43 +1,42 @@
...
-        NSDictionary* lastVersion = [preferences objectForKey:LastVersionPreference];
-        if (lastVersion) {
-            lastVersionLabelOrAppVersionParameter = [CodePushReportingManager convertStringParameter:[lastVersion objectForKey:LastVersionLabelOrAppVersionKey]];
-            lastVersionDeploymentKeyParameter = [CodePushReportingManager convertStringParameter:[lastVersion objectForKey:LastVersionDeploymentKeyKey]];
+        NSString* labelParameter = [CodePushReportingManager convertStringParameter:statusReport.label];
+        NSString* appVersionParameter = [CodePushReportingManager convertStringParameter:statusReport.appVersion];
+        NSString* deploymentKeyParameter = [CodePushReportingManager convertStringParameter:statusReport.deploymentKey];
+        NSString* lastVersionLabelOrAppVersionParameter = statusReport.lastVersionLabelOrAppVersion;
+        NSString* lastVersionDeploymentKeyParameter = statusReport.lastVersionDeploymentKey;
+        if (!lastVersionLabelOrAppVersionParameter || !lastVersionDeploymentKeyParameter) {
```
scottbommarito: this is doing even more shady "cover-up" kind of stuff. although u can assume that, given your constructor, if one of these is null the other is also probably null as well, but if there's a bug and only one is null now it's harder to tell where the issue is because you replace both anyway

src/android/CodePushReportingManager.java
geof90: Doesn't work. You can return the result of a value assignment.

scottbommarito: really? I thought u could haha

geof90: screen shot 2016-05-13 at 11 59 16 am

scottbommarito: rip

src/android/CodePush.java
scottbommarito: (basically, I'm just wondering why this `else if` isn't also attached to the other `if (pendingInstall...` block above as well)

src/android/CodePushReportingManager.java
geof90: So it turns out it actually works! Thanks. I stupidly tried it on a void method and that's why it didn't work initially. However, I want to keep the original code cuz its more readable, if that's ok with you.

scottbommarito: nice! feel free to keep the original code, what matters is that we learned something about java!

src/android/CodePush.java
geof90: `switch` on a string is only added in Java 7, which will not work on older versions of Android.

src/android/CodePush.java
geof90: I think I'll rename it to `execNotifyApplicationReady`. That makes JS and native use the same name which I think is way simpler. The RN implementation does that too.

scottbommarito: perfect!

src/android/CodePush.java
geof90: On restart, notifyApplicationReady will be called anyway, and that will handle the reporting.

src/android/CodePushReportingManager.java
geof90: This is the only place, I will keep that in mind the next time we do a similar check. (I hate that Java doesn't have support for comparisons like this in `java.lang`)

src/android/CodePushReportingManager.java
geof90: The way to think about it is, it's null by default, and by default we should always be going to the "last version" stored in codePushPreferences. 

The only time we choose not to use it if it's explicitly set by the caller of this method, e.g. for failed reports where we want to send the exact same data that was previously sent.

scottbommarito: oh I see...
(this is the kind of thing i'd add a comment for btw)

src/android/CodePushReportingManager.java
geof90: Checking prefs means reading data from disk, which could be a fairly long operation and I prefer not to do that everytime the app resumes. We don't want to be responsible for slow performance on apps using our plugin.

src/android/StatusReport.java
geof90: Why?

src/android/CodePushReportingManager.java
scottbommarito: http://stackoverflow.com/questions/19148282/read-speed-of-sharedpreferences

they're in memory after the first read/write, so I wouldn't be worried about performance. ultimately it's fine as a variable, but if this behavior gets more complex I would definitely switch to using the prefs every time.

src/android/StatusReport.java
geof90: I prefer for logic like that to be handled in the controller (i.e. the CodePushReportingManager), not the model class. Arguably, its more OO.

src/android/CodePushReportingManager.java
geof90: Even if it's cached after the first read, having this Boolean still saves the first read in cases where we set the failed report and the process doesn't get killed.

src/ios/CodePushReportingManager.m
geof90: Would changing it to `&&` be better?

src/android/CodePush.java
scottbommarito: but doesn't `notifyApplicationReady` get called by a resume update as soon as it starts as well?

geof90: We want to check for failed reports everytime the app resumes, not just when there's an update.

src/android/StatusReport.java
scottbommarito: consistency...but that's fine haha

src/android/CodePushReportingManager.java
geof90: What would the comment be?
"if not explictly set, use the value of the last version stored in preferences." seems like stating the obvious to me.

src/ios/CodePushReportingManager.m
scottbommarito: yes

src/android/CodePush.java
scottbommarito: oh I see, thanks

src/android/CodePushReportingManager.java
scottbommarito: well yes that's obvious, but what's not obvious is that they exist as a member variable on the object in the first place because if it is specified, it is meant to refer to a failed update, and if it is not, it is the first time attempting to report the update. I would put it in the status report class tho, not here

src/android/CodePushReportingManager.java
scottbommarito: oh I see, you're using a different shared preferences for each preference...I don't think this is necessary, I think you can put all the preferences in a single shared preference and it will be faster, but given that all the preferences are separate, this storing this variable like this is definitely faster.

scottbommarito: (i'll make a pr with them all in one file after u push this, I have plenty of time while waiting for react-native tests lol)

scottbommarito:
LGTM

PR #122, opened 2016-05-19
Fix rollback reporting

geof90:
This PR fixes an issue with reporting metrics in the case of a rollback.

Originally, `reportStatus` was called in the `handleUnconfirmedInstall` method, however, during that time the app startup process, the page and the `CodePush` javascript plugin has not been fully initialized, so executing the JS `codePush.reportStatus(...)` would just no-op.

This PR addresses that by caching the rollback status report in memory during `handleUnconfirmedInstall`, and only report the rollback when `notifyApplicationReady()` is called, at which point the app would be in the `deviceReady` state and all the plugins would have been fully initialized.


dtivel:
Looks good.

ryuyu:
:shipit:

shishirx34:
:shipit:

PR #125, opened 2016-05-23
Fix #124

geof90:
This PR fixes #124 by ensuring that we only backup the current package if it is not still pending.
Added test coverage for this scenario.

PR #127, opened 2016-05-24
fixed the test in #125

scottbommarito:
i'm probably gonna have to make some more changes so that the tests are easier to understand (eventually) but this fixes Geoffrey's test case that was failing on Android only because the emulator is slower than the iOS emulator

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -862,54 +862,56 @@ function runTests(targetPlatform: platform.IPlatform, useWkWebView: boolean): vo
             }
         });
 
-        describe("#localPackage.installOnNextRestart2x", function() {
+        if (!onlyRunCoreTests) {
+            describe("#localPackage.installOnNextRestart2x", function() {
 
-            afterEach(() => {
-                cleanupTest();
+                afterEach(() => {
+                    cleanupTest();
+                });
+
+                before(() => {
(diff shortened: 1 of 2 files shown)


geof90:
LGTM!

scottbommarito:
@lostintangent 
i think i need another sign-off lol

scottbommarito:
had to make one more change because the test was not labeled as not a core test

(yea ik it's confusing, will be much clearer soon)

perfect score on VSTS
https://mobiledevops.visualstudio.com/VS%20Client/_build?_a=summary&buildId=4300

test/test.ts
```diff
@@ -862,59 +862,56 @@ function runTests(targetPlatform: platform.IPlatform, useWkWebView: boolean): vo
             }
         });
 
-        describe("#localPackage.installOnNextRestart2x", function() {
+        if (!onlyRunCoreTests) {
```
dlebu: Mocha has the ability to filter tests after name. If all core tests would contain a `core` prefix, you would not need to do this.

dlebu:
LGTM

test/test.ts
scottbommarito: interesting...i'm changing the whole implementation pretty soon (with the change that will support react-native testing) but this is good to know

PR #128, opened 2016-05-24
Fix #126

### LINKED ISSUE #126: Application crash on Android at startup
We are sometimes seeing an exception at startup using CodePush, then configuration is basic and we are just using .sync() to deal with everything at device ready.

This was seen when deploying a new version of the app via VS-TACO after already having an earlier build of the app on the device which had already updated itself successfully via CodePush.

The earlier build used exactly the same plugins and had the same binary version, all the changes were in the '.\www' folder. 

This is using CodePush CLI v1.11.0-beta and Cordova CLI 6.0.0

The trace is as follows:

D/AndroidRuntime(13342): Shutting down VM

E/AndroidRuntime(13342): FATAL EXCEPTION: main

E/AndroidRuntime(13342): Process: com.emrgroup.testapp, PID: 13342

E/AndroidRuntime(13342): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.emrgroup.testapp/com.emrgroup.testapp.MainActivity}: java.lang.NullPointerException: Attempt to read from field 'java.lang.String com.microsoft.cordova.CodePushPackageMetadata.label' on a null object reference

E/AndroidRuntime(13342): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2808)

E/AndroidRuntime(13342): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2873)

E/AndroidRuntime(13342): 	at android.app.ActivityThread.access$900(ActivityThread.java:181)

E/AndroidRuntime(13342): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1482)

E/AndroidRuntime(13342): 	at android.os.Handler.dispatchMessage(Handler.java:102)

E/AndroidRuntime(13342): 	at android.os.Looper.loop(Looper.java:145)

E/AndroidRuntime(13342): 	at android.app.ActivityThread.main(ActivityThread.java:6145)

E/AndroidRuntime(13342): 	at java.lang.reflect.Method.invoke(Native Method)

E/AndroidRuntime(13342): 	at java.lang.reflect.Method.invoke(Method.java:372)

E/AndroidRuntime(13342): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)

E/AndroidRuntime(13342): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)

E/AndroidRuntime(13342): Caused by: java.lang.NullPointerException: Attempt to read from field 'java.lang.String com.microsoft.cordova.CodePushPackageMetadata.label' on a null object reference

E/AndroidRuntime(13342): 	at com.microsoft.cordova.CodePush.handleUnconfirmedInstall(CodePush.java:333)

E/AndroidRuntime(13342): 	at com.microsoft.cordova.CodePush.onStart(CodePush.java:446)

E/AndroidRuntime(13342): 	at org.apache.cordova.PluginManager.onStart(PluginManager.java:276)

E/AndroidRuntime(13342): 	at org.apache.cordova.CordovaWebViewImpl.handleStart(CordovaWebViewImpl.java:462)

E/AndroidRuntime(13342): 	at org.apache.cordova.CordovaActivity.onStart(CordovaActivity.java:293)

E/AndroidRuntime(13342): 	at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1234)

E/AndroidRuntime(13342): 	at android.app.Activity.performStart(Activity.java:6412)

E/AndroidRuntime(13342): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2771)

E/AndroidRuntime(13342): 	... 10 more

geof90:
Couple of changes in this PR: 

- Renamed `codePushPackageManager.xxFirstRun()` methods to `codePushPackageManager.xxBinaryFirstRun()` to better reflect what those methods do and prevent confusion with the `isFirstRun` flag on our update metadata.

- Removed the call to `reportStatus` in app start (previously in `handleAppStart`), because it was causing crashes on Android whenever a rollback occurs on app start. Instead, delayed the `reportStatus` calls to `notifyApplicationReady` using the `xxBinaryFirstRun()` mechanisms already in place, that way its all the `reportStatus` calls are now in one place.

- Split up the `handleAppStart` method into `clearDeploymentsIfBinaryUpdated` and `navigateToLocalDeploymentIfExists` methods. This is needed in order to fix #126, which causes a crash when `handleUnconfirmedInstall` tries to rollback a "pending" update which unknowingly got cleared by `handleAppStart`. By splitting it up, we now can always call `clearDeploymentsIfBinaryUpdated` before all the other app initialization logic, which prevents us from getting "pending" information that is actually outdated and needs to be cleared.

- Added logic in JS `reportStatus` to no-op if the appVersion and label of the new update report is identical with the old update, which could happen whenever a new binary is deployed directly to the device, or when another app deployment service like HockeyApp updates the app without bumping the binary version.

### DIFF
--- a/bin/www/codePush.js
+++ b/bin/www/codePush.js
@@ -25,6 +25,11 @@ var CodePush = (function () {
         cordova.exec(installSuccess, errorCallback, "CodePush", "restartApplication", []);
     };
     CodePush.prototype.reportStatus = function (status, label, appVersion, deploymentKey, previousLabelOrAppVersion, previousDeploymentKey) {
+        if ((!label || label === previousLabelOrAppVersion) &&
+            appVersion === previousLabelOrAppVersion &&
+            deploymentKey == previousDeploymentKey) {
+            return;
+        }
         var createPackageForReporting = function (label, appVersion) {
             return {
                 label: label, appVersion: appVersion, deploymentKey: deploymentKey,
--- a/bin/www/codePush.js
+++ b/bin/www/codePush.js
@@ -27,7 +27,7 @@ var CodePush = (function () {
     CodePush.prototype.reportStatus = function (status, label, appVersion, deploymentKey, previousLabelOrAppVersion, previousDeploymentKey) {
         if ((!label || label === previousLabelOrAppVersion) &&
             appVersion === previousLabelOrAppVersion &&
-            deploymentKey == previousDeploymentKey) {
+            deploymentKey === previousDeploymentKey) {
             return;
         }
         var createPackageForReporting = function (label, appVersion) {
--- a/bin/www/codePush.js
+++ b/bin/www/codePush.js
@@ -25,9 +25,8 @@ var CodePush = (function () {
         cordova.exec(installSuccess, errorCallback, "CodePush", "restartApplication", []);
     };
     CodePush.prototype.reportStatus = function (status, label, appVersion, deploymentKey, previousLabelOrAppVersion, previousDeploymentKey) {
-        if ((!label || label === previousLabelOrAppVersion) &&
-            appVersion === previousLabelOrAppVersion &&
-            deploymentKey === previousDeploymentKey) {
+        if (((!label && appVersion === previousLabelOrAppVersion) || label === previousLabelOrAppVersion)
+            && deploymentKey === previousDeploymentKey) {
             return;
         }
         var createPackageForReporting = function (label, appVersion) {
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -278,7 +278,7 @@ public class CodePush extends CordovaPlugin {
             long nativeBuildTime = Utilities.getApkEntryBuildTime(RESOURCES_BUNDLE, this.cordova.getActivity());
             if (nativeBuildTime != -1) {
                 String currentAppTimeStamp = String.valueOf(nativeBuildTime);
-                if (deployedPackageTimeStamp != null && !deployedPackageTimeStamp.equals(currentAppTimeStamp)) {
+                if (!currentAppTimeStamp.equals(deployedPackageTimeStamp)) {
                     this.codePushPackageManager.cleanDeployments();
                     this.codePushPackageManager.clearFailedUpdates();
                     this.codePushPackageManager.clearPendingInstall();
@@ -298,7 +298,7 @@ public class CodePush extends CordovaPlugin {
                 try {
                     navigateToFile(startPage);
                 } catch (MalformedURLException e) {
-                /* empty - if there is an exception, the app will launch with the bundled content */
+                    /* empty - if there is an exception, the app will launch with the bundled content */
                 }
             }
         }
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -278,14 +278,12 @@ public class CodePush extends CordovaPlugin {
             long nativeBuildTime = Utilities.getApkEntryBuildTime(RESOURCES_BUNDLE, this.cordova.getActivity());
             if (nativeBuildTime != -1) {
                 String currentAppTimeStamp = String.valueOf(nativeBuildTime);
-                if ((deployedPackageTimeStamp != null) && (currentAppTimeStamp != null)) {
-                    if (!deployedPackageTimeStamp.equals(currentAppTimeStamp)) {
-                        this.codePushPackageManager.cleanDeployments();
-                        this.codePushPackageManager.clearFailedUpdates();
-                        this.codePushPackageManager.clearPendingInstall();
-                        this.codePushPackageManager.clearInstallNeedsConfirmation();
-                        this.codePushPackageManager.clearBinaryFirstRunFlag();
-                    }
+                if (deployedPackageTimeStamp != null && !deployedPackageTimeStamp.equals(currentAppTimeStamp)) {
+                    this.codePushPackageManager.cleanDeployments();
+                    this.codePushPackageManager.clearFailedUpdates();
+                    this.codePushPackageManager.clearPendingInstall();
+                    this.codePushPackageManager.clearInstallNeedsConfirmation();
+                    this.codePushPackageManager.clearBinaryFirstRunFlag();
                 }
             }
         }
@@ -293,16 +291,14 @@ public class CodePush extends CordovaPlugin {
 
     private void navigateToLocalDeploymentIfExists() {
         CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
-        if (deployedPackageMetadata != null) {
-            if (deployedPackageMetadata.localPath != null) {
-                File startPage = this.getStartPageForPackage(deployedPackageMetadata.localPath);
-                if (startPage != null) {
-                    /* file exists */
-                    try {
-                        navigateToFile(startPage);
-                    } catch (MalformedURLException e) {
-                    /* empty - if there is an exception, the app will launch with the bundled content */
-                    }
+        if (deployedPackageMetadata != null && deployedPackageMetadata.localPath != null) {
+            File startPage = this.getStartPageForPackage(deployedPackageMetadata.localPath);
+            if (startPage != null) {
+                /* file exists */
+                try {
+                    navigateToFile(startPage);
+                } catch (MalformedURLException e) {
+                /* empty - if there is an exception, the app will launch with the bundled content */
                 }
             }
         }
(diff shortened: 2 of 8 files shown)


www/codePush.ts
```diff
@@ -63,6 +63,14 @@ class CodePush implements CodePushCordovaPlugin {
      * !!! This function is called from the native side, please make changes accordingly. !!!
      */
     public reportStatus(status: number, label: string, appVersion: string, deploymentKey: string, previousLabelOrAppVersion?: string, previousDeploymentKey?: string) {
+       if ((!label || label === previousLabelOrAppVersion) &&
+           appVersion === previousLabelOrAppVersion &&
+           deploymentKey == previousDeploymentKey) {
```
scottbommarito: is it correct that this equality check is `==` instead of `===`?

src/android/CodePush.java
```diff
@@ -270,6 +270,44 @@ private void cleanOldPackageSilently() {
...
 
+    private void clearDeploymentsIfBinaryUpdated() {
+        /* check if we have a deployed package already */
+        CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
+        if (deployedPackageMetadata != null) {
+            String deployedPackageTimeStamp = deployedPackageMetadata.nativeBuildTime;
+            long nativeBuildTime = Utilities.getApkEntryBuildTime(RESOURCES_BUNDLE, this.cordova.getActivity());
+            if (nativeBuildTime != -1) {
+                String currentAppTimeStamp = String.valueOf(nativeBuildTime);
+                if ((deployedPackageTimeStamp != null) && (currentAppTimeStamp != null)) {
```
scottbommarito: u can compress this and the following if with
`if ((deployedPackageTimeStamp != null) && !deployedPackageTimeStamp.equals(currentAppTimeStamp))`
afaik `.equals` returns null if the thing in the parentheses is null.

src/android/CodePush.java
```diff
@@ -270,6 +270,44 @@ private void cleanOldPackageSilently() {
...
+                        this.codePushPackageManager.clearBinaryFirstRunFlag();
+                    }
+                }
+            }
+        }
+    }
+
+    private void navigateToLocalDeploymentIfExists() {
+        CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
+        if (deployedPackageMetadata != null) {
```
scottbommarito: you can compress these `if`s with
`if (deployedPackageMetadata != null && deployedPackageMetadata.localPath != null)`

src/android/CodePush.java
geof90: I was keeping it as similar to what it was before as possible... but okay.

src/ios/CodePush.m
```diff
@@ -224,45 +224,45 @@ - (void)dealloc {
...
         if (deployedPackageNativeBuildTime != nil && applicationBuildTime != nil) {
-            if ([deployedPackageNativeBuildTime isEqualToString: applicationBuildTime] ) {
-                // same version, safe to launch from local storage
-                if (deployedPackageMetadata.localPath) {
-                    [self redirectStartPageToURL: deployedPackageMetadata.localPath];
-                }
-            }
-            else {
-                // installed native version is different from package version => clean up deployed package and do not modify start page
+            if (![deployedPackageNativeBuildTime isEqualToString: applicationBuildTime]) {
```
scottbommarito: (i don't know if Objective-C's `isEqualToString` has the same behavior as Java's `.equals` but i would look into it)

src/ios/CodePush.m
```diff
@@ -224,45 +224,45 @@ - (void)dealloc {
...
+                // installed native version is different from package version
```
scottbommarito: `// package version is incompatible with installed native version`? technically they're always supposed to be different and the issue here is that they're not compatible, not that they're different.

src/android/CodePush.java
scottbommarito: also i noticed you added a comment in the iOS native code around here describing that this implies the package and native version are incompatible. i don't know if it's necessary but if you're going to keep it i'd put it here as well for consistency's sake

src/ios/CodePush.m
```diff
@@ -224,45 +224,45 @@ - (void)dealloc {
...
-                StatusReport* statusReport = [[StatusReport alloc] initWithStatus:STORE_VERSION andLabel:nil andAppVersion:appVersion andDeploymentKey:deploymentKey];
-                [CodePushReportingManager reportStatus:statusReport withWebView:self.webView];
+                [CodePushPackageManager clearBinaryFirstRunFlag];
             }
         }
     }
 }
 
+- (void)navigateToLocalDeploymentIfExists {
+    // check if we have a deployed package
```
scottbommarito: u may not hear this from me that much but this comment is kinda redundant LOL

src/android/CodePush.java
```diff
@@ -270,6 +270,44 @@ private void cleanOldPackageSilently() {
...
+            if (deployedPackageMetadata.localPath != null) {
+                File startPage = this.getStartPageForPackage(deployedPackageMetadata.localPath);
+                if (startPage != null) {
+                    /* file exists */
+                    try {
+                        navigateToFile(startPage);
+                    } catch (MalformedURLException e) {
+                    /* empty - if there is an exception, the app will launch with the bundled content */
```
scottbommarito: tab this comment in to be aligned with where code would be

scottbommarito:
i hope u aren't too annoyed by my nit-pick-a-palooza

`// LGTM`

src/android/CodePush.java
```diff
@@ -270,6 +270,40 @@ private void cleanOldPackageSilently() {
...
+    private void navigateToLocalDeploymentIfExists() {
+        CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
+        if (deployedPackageMetadata != null && deployedPackageMetadata.localPath != null) {
+            File startPage = this.getStartPageForPackage(deployedPackageMetadata.localPath);
+            if (startPage != null) {
+                /* file exists */
+                try {
+                    navigateToFile(startPage);
+                } catch (MalformedURLException e) {
+                /* empty - if there is an exception, the app will launch with the bundled content */
```
scottbommarito: (a commit u did while i was reviewing got rid of a pr comment i added, but this comment should be tabbed in by one)

bin/www/codePush.js
```diff
@@ -25,6 +25,11 @@ var CodePush = (function () {
         cordova.exec(installSuccess, errorCallback, "CodePush", "restartApplication", []);
     };
     CodePush.prototype.reportStatus = function (status, label, appVersion, deploymentKey, previousLabelOrAppVersion, previousDeploymentKey) {
+        if ((!label || label === previousLabelOrAppVersion) &&
+            appVersion === previousLabelOrAppVersion &&
+            deploymentKey === previousDeploymentKey) {
```
shishirx34: shouldn't this be: 
`((!label && appVersion === previousLabelOrAppVersion) || (label === previousLabelOrAppVersion)) && deploymentKey === previousDeploymentKey`

ryuyu:
LGTM

scottbommarito:
WAIT A MINUTE we should probably make a test case for this right

www/codePush.ts
```diff
@@ -63,6 +63,14 @@ class CodePush implements CodePushCordovaPlugin {
      * !!! This function is called from the native side, please make changes accordingly. !!!
      */
     public reportStatus(status: number, label: string, appVersion: string, deploymentKey: string, previousLabelOrAppVersion?: string, previousDeploymentKey?: string) {
+       if ((!label || label === previousLabelOrAppVersion) &&
+           appVersion === previousLabelOrAppVersion &&
+           deploymentKey === previousDeploymentKey) {
```
shishirx34: same comment as above...

Shouldn't you check either `label` or `appVersion` against `previousLabelOrAppVersion` and not both? If `label` is defined and is equal to `previousLabelOrAppVersion` and `appVersion` which I am assuming is not same as `label` since it can't be, in that case it is not going to be equal to `previousLabelOrAppVersion` then it should return, shouldn't it?

bin/www/codePush.js
geof90: aw right.. good catch.

shishirx34:
LGTM

PR #135, opened 2016-06-13
fix device list parsing issue

scottbommarito:
the parsing was including the device pairs as well as the regular devices, which would cause the booting of the simulator to fail and thus failing our automated tests occasionally.

the plugin testing framework has been updated with this change, but this is a hotfix until OSS approves it.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -217,7 +217,6 @@ gulp.task("tslint", function () {
 });
 
 gulp.task("clean", function () {
-    var del = require("del");
     return del([binPath + "/**"], { force: true });
 });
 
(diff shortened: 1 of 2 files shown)


scottbommarito:
@geof90 @lostintangent 
hoping for a quick approval so i can fix the tests

gulpfile.js
```diff
@@ -217,7 +217,6 @@ gulp.task("tslint", function () {
 });
 
 gulp.task("clean", function () {
-    var del = require("del");
```
scottbommarito: (this is unrelated to the device list parsing, but i noticed this was redeclared after already being declared at the top of the file so i got rid of this)

geof90:
LGTM

scottbommarito:
@ryuyu @shishirx34 @lostintangent

shishirx34:
LGTM

PR #136, opened 2016-06-13
getAcquisitionManager: specific error if no version in config.xml

ajoslin:
I'm generating a config.xml in my project for each environment, and it turns out my config.xml generation script wasn't specifying the version correctly.

I wasted a lot of time today because of the vague error that was thrown when appVersion isn't found.

Hopefully this more specific error will save someone else some time.

### DIFF
--- a/www/sdk.ts
+++ b/www/sdk.ts
@@ -44,7 +44,9 @@ class Sdk {
             NativeAppInfo.getServerURL((serverError: Error, serverURL: string) => {
                 NativeAppInfo.getDeploymentKey((depolymentKeyError: Error, deploymentKey: string) => {
                     NativeAppInfo.getApplicationVersion((appVersionError: Error, appVersion: string) => {
-                        if (!serverURL || !appVersion) {
+                        if (!appVersion) {
+                            callback(new Error("Could not get the app version. Please check your config.xml file."), null);
+                        } else if (!serverURL) {
                             callback(new Error("Could not get the CodePush configuration. Please check your config.xml file."), null);
                         } else {
                             Sdk.DefaultConfiguration = {


lostintangent:
Apologies for the confusion! We haven't heard any issues here yet, so perhaps your dynamic config.xml generation is somewhat unique. Thanks so much for the contribution!

Could you actually share a little more detail about your specific issue? We don't actually read the app version from the config.xml at runtime, so I'm curious how your app would have been deployed without having a version we could retrieve for it.

ajoslin:
Sure.

My build step passes a file `config.template.xml` through an underscore template to spit out a `config.xml` depending on the environment (development/staging/production). This template has variables for:

- Version (matching pkg.version)
- Bundle identifier 
- App name
- code push key (empty in development, set in staging/production)

I was doing a build against staging, then running `cordova run ios`. When I started the app on the phone, it did a version check it gave the generic error: `Could not get the CodePush configuration. Please check your config.xml file.`

First, I checked that my CodePush keys were *really* being set in many different ways, even going as far as to manually pass them in via Javascript.

After realizing that wasn't the problem, I followed the stacktrace of the error and found out it was because of version being null, and I realized that my config.xml had been generated with `version=""`.

Does that make sense?

lostintangent:
Makes sense, thanks a lot for explaining that! I'm surprised that running "cordova run ios" didn't report an error in this scenario, since the version attribute is supposed to be required by Cordova, and the resulting Info.plist property is definitely required in order to actually publish your app to the store. 

Your change looks good to me!

scottbommarito:
looks good to me as well! nice fix!

ajoslin:
:+1: Thanks for the great tool.

PR #141, opened 2016-06-16
Fixing header (iOS)

lostintangent:
The `updateSuccess` method was removed from the `CodePush` class in `v1.8.0`, and this PR simply removes it from the corresponding header file so that Xcode doesn't display a warning about it not being implemented.


scottbommarito:
`// LGTM`

geof90:
LGTM

PR #142, opened 2016-06-16
Backgrounding expensive operations (iOS)

lostintangent:
This addresses https://github.com/Microsoft/code-push/issues/256 by ensuring that expensive native operations are offloaded to a background thread. I modified the Cordova warning threshold to 1ms (instead of 10) for testing purposes, and moved all methods that took longer than that off of the UI thread. The biggest contributors were `getBinaryHash` and `notifyApplicationReady`, both of which were ~15ms+, which definitely shouldn't be done on the UI thread. The `preInstall` and `install` methods were steadily around 2ms as well, so I decided to offload them even though they aren't as impactful. All other methods were sub-millisecond, so I decided not to go crazy with this PR (even though our React Native plugin runs all native code on a background thread). We can always improve this later as needed.

The churn below is as a result of the "wrapping" that the background thread code introduces. The basic structure of my changes are the following (based on [Cordova guidelines](https://cordova.apache.org/docs/en/latest/guide/platforms/ios/plugin.html#threading)), and impacts the aforementioned four methods:

```objective-c
[self.commandDelegate runInBackground:^{
    // Existing code here
}];
```
Note that I didn't look into Android at all, since I wanted to address the user-reported iOS issue specifically first. We can investigate Android optimization opportunities separately.

### DIFF
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -19,25 +19,27 @@ NSString* const DeploymentKeyPreference = @"codepushdeploymentkey";
 StatusReport* rollbackStatusReport = nil;
 
 - (void)getBinaryHash:(CDVInvokedUrlCommand *)command {
-    CDVPluginResult* pluginResult = nil;
-    NSString* binaryHash = [CodePushPackageManager getCachedBinaryHash];
-    if (binaryHash) {
-        pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK
-                                         messageAsString:binaryHash];
-    } else {
-        NSError* error;
-        binaryHash = [UpdateHashUtils getBinaryHash:&error];
-        if (error) {
-            pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_ERROR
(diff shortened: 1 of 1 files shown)


scottbommarito:
`// LGTM` nice fix! i've definitely noticed the error message on android but never looked into it, shouldn't be any harder than it was to do this. all you have to do is make a runnable and put the code in there and then add it to a threadpool if i remember correctly

geof90:
LGTM!

geof90:
Its interesting why notifyApplicationReady would take that long.

geof90:
Just eyeballing by the number of IO operations, `clearDeploymentsIfBinaryUpdated` might be expensive too, although it is a rare operation.

lostintangent:
@geof90 I considered doing that, but I wasn't sure whether that could introduced a race condition since the `pluginInitialize` method currently clears the deployments synchronously and then attempts to redirect the WebView. The later operation seems dependent on the `clearDeploymentsIfBinaryUpdated` operation completing, and therefore, I'm not sure if we can simply offload it to a background thread.

For this PR, I was trying to optimize our JS->native calls, but I think we can separately look into how to optimize our app start (namely the `pluginInitialize` method).

PR #143, opened 2016-06-17
WKWebView fix

### LINKED ISSUE #137: Install does not complete update from resume
Hi,

I'm having trouble getting my app to update on resume, I can see in the logs that the update is successfully identified and downloaded. The last SyncStatus I see is INSTALLING_UPDATE before some sandbox deny errors, one of which seems to be for reading the index.html file in the `codedeploy/versions` folder. Installing on restart does seem to work though.

I've setup my `config.xml` as per the instructions [here](https://github.com/Microsoft/cordova-plugin-code-push#getting-started) but maybe there is something else I need to configure as I'm using the WKWebView provided by `cordova-plugin-wkwebview-engine`?

Any help would be greatly appreciated.

Cheers

lostintangent:
This fixes #137 by using the `WebViewEngine` to navigate the WebView instead of using the deprecated `loadRequest:` method on the UI/WKWebView class (which enforces new security rules in iOS 9.3+). The `webViewEngine` property was only added in Cordova iOS 4.0.0, and our plugin currently supports 3.9.0+, which is why I'm using `performSelector`/`id` type to actually perform the navigation. At some point, it might make sense for us to just start requiring 4.0.0+, at which point we can clean this code up. For now though, I don't want to change our platform compatibility.

Additionally, this fixes a regression I caused with #142. I had only tested that fix with `WKWebView` and it turns out that the `UIWebView` requires you to execute JS from the UI thread, so I need to ensure that we dispatch that call when the app is using the `UIWebView`.

I tested this thoroughly with Cordova iOS 3.9.0, and Cordova iOS 4.1.1, with both the `UIWebView` and `WKWebView` and everything works beautifully.

### DIFF
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -311,9 +311,12 @@ StatusReport* rollbackStatusReport = nil;
 }
 
 - (void)loadURL:(NSURL*)url {
-    // Cast to a UIWebView here to enable call to loadRequest.
-    // Both UIWebView and WKWebview share this call, so regardless of which type the returned webview is, this will work.
-    [(UIWebView*)self.webView loadRequest:[NSURLRequest requestWithURL:url]];
+    if ([self respondsToSelector:@selector(webViewEngine)]) {
+        id webViewEngine = [self performSelector:@selector(webViewEngine)];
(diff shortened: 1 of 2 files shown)


shishirx34:
Looks good.

scottbommarito:
`// LGTM`

geof90:
LGTM!

PR #145, opened 2016-06-18
WKWebView fix

lostintangent:
This is the final piece of the fix for #147.

### DIFF
--- a/src/ios/CDVWKWebViewEngine+CodePush.m
+++ b/src/ios/CDVWKWebViewEngine+CodePush.m
@@ -1,0 +1,38 @@
+#if defined(__has_include)
+#if __has_include("CDVWKWebViewEngine.h")
+#if !(TARGET_OS_SIMULATOR)
+
+#import "CDVWKWebViewEngine.h"
+
+@implementation CDVWKWebViewEngine (CodePush)
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
+
+- (id)loadRequest:(NSURLRequest *)request {
+    NSURL *readAccessURL;
+    
+    // If the app is attempting to load a CodePush update, then we can lock the WebView down to
+    // just the CodePush "versions" directory. This prevents non-CodePush assets from being accessible,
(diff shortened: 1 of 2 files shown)


src/ios/CDVWKWebViewEngine+CodePush.m
```diff
@@ -0,0 +1,38 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
+
+- (id)loadRequest:(NSURLRequest *)request {
+    NSURL *readAccessURL;
+    
+    // If the app is attempting to load a CodePush update, then we can lock the WebView down to
+    // just the CodePush "versions" directory. This prevents non-CodePush assets from being accessible,
+    // while stll allowing us to navigate to a future update, as well as to the binary if a rollback is needed.
+    if ([request.URL.absoluteString containsString:@"codepush"]) {
```
geof90: If I don't recall wrongly, `containsString` will not work on iOS 7

src/ios/CDVWKWebViewEngine+CodePush.m
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
+
+- (id)loadRequest:(NSURLRequest *)request {
+    NSURL *readAccessURL;
+    
+    // If the app is attempting to load a CodePush update, then we can lock the WebView down to
+    // just the CodePush "versions" directory. This prevents non-CodePush assets from being accessible,
+    // while stll allowing us to navigate to a future update, as well as to the binary if a rollback is needed.
```
geof90: still

geof90:
LGTM!

ryuyu:
Lgtm

src/ios/CDVWKWebViewEngine+CodePush.m
lostintangent: I confirmed that the WKWebViewEngine is only "installed" on iOS9, and therefore, this code won't be called on iOS 7, even if the app has the WKWebView plugin installed. Thanks for catching this though, since I admittedly hadn't thought about back company with a wrote this.

PR #147, opened 2016-06-24
Fixing PhoneGap Build issue

lostintangent:
This PR fixes https://github.com/Microsoft/code-push/issues/268 by removing the runtime resolution of the `loadRequest:` selector, which can in some circumstances be ambitious due to their being multiple "versions" of this method available. Instead, we can perform the `WebViewEngine` vs. `WebView` check at compile-time, based on whether or not the host app is running Cordova iOS 4.0.0+, and then call the right `loadRequest:` method directly on each type, so that we don't need dynamic runtime resolution anyways.


ryuyu:
LGTM

geof90:
LGTM!

PR #167, opened 2016-09-19
Add `cordovaDependencies` section to package.json

vladimir-kotikov:
This PR adds `engines` element to package.json based on [plugin documentation](https://github.com/Microsoft/cordova-plugin-code-push/blob/master/README.md#supported-cordova-platforms) to comply w/ [new cordova plugin fetching model](https://github.com/cordova/cordova-discuss/blob/master/proposals/plugin-version-fetching.md).

This also adds a 'protective' entry for next major plugin version to protect end-users from fetching edge versions of the plugin by incompatible version of cordova. See corresponding [discussion on mailing list](http://apache.markmail.org/thread/p73loqtvbzvfzsv5) for more details and reasons behind this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,18 @@
       "ios"
     ]
   },
+  "engines": {
+    "cordovaDependencies": {
+      "1.8.x": {
+        "cordova": ">=5.0.0",
+        "cordova-android": ">=4.0.0",
+        "cordova-ios": ">=3.9.0"
+      },
+      "2.0.0": {
+        "cordova": ">=100"
+      }
+    }
+  },
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
@silhouettes , thanks for looking into this!

Regarding dependencies to other plugins - this is a separate thing. Cordova does not support adding dependencies to other plugins via `package.json` so you still need to keep these entries in `plugin.xml`

silhouettes:
Sounds good to me! The only reason I asked is because of this section in the page you linked:

```
<constraint> allowed values:

* cordova
* Platforms (e.g. cordova-android)
* Plugins (e.g. cordova-plugin-camera)
```

Issue #196, opened 2016-12-07
Android - New Binary Version Issues


TombolaShepless:
When I create a new binary version, Code Push still uses the last package stored in the apps data and doesn't load the new binary versions assets. These are the steps taken:

1) Build and run v1.0.0 (fresh APK install on the device)
`cordova run android`

2) Create new code push for v1.0.0 (label v1)
`code-push release-cordova test android --deploymentName dev`

3) Download and install code push (label v1)

4) Create new code push for v1.0.1 (label v1, to test binary version restrictions)
`Manually bump config.xml to 1.0.1`
`code-push release-cordova test android --deploymentName dev`

5) Open app and ensure v1.0.1 (label v1) is NOT installed (this works as expected)

6) Create v1.0.1 APK and install on the device
`cordova run android`

7) Open app (should now be v1.0.1)

8) App is actually v1.0.1, but is loading assets from last successful code push (v1.0.0 label v1).

Is there something I'm doing wrong here? Or is this process dependent on the Play Store update process?

Any help is appreciated.

Shepless:
Ok after a bit of digging it looks like [this](https://github.com/Microsoft/cordova-plugin-code-push/blob/adb82ddc2dfa1cf03128eb4db8137cce081c3798/src/android/CodePush.java#L273) is the problem. More specifically, [this](https://github.com/Microsoft/cordova-plugin-code-push/blob/adb82ddc2dfa1cf03128eb4db8137cce081c3798/src/android/Utilities.java#L65) is always returning a value that matches `deployedPackageTimeStamp`. See [this](https://github.com/Microsoft/cordova-plugin-code-push/blob/adb82ddc2dfa1cf03128eb4db8137cce081c3798/src/android/CodePush.java#L281) line. 

I'm not savvy with Android native development but [this](https://developer.android.com/reference/android/content/pm/PackageInfo.html#lastUpdateTime) looks like a more appropriate API for this task? It's been in the API since version 9 so support wouldn't be an issue. I guess theres a reason you guys are doing the way you are, I'm just not aware of it.

TombolaShepless:
[...]

PR #201, opened 2016-12-20
Getting correct build time for apk

matrosov-nikita:
We should use `classes.dex` instead of `resources.arsc` to get correct build time for apk (https://github.com/Microsoft/cordova-plugin-code-push/issues/196)

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -23,7 +23,7 @@ import java.util.Date;
  */
 public class CodePush extends CordovaPlugin {
 
-    public static final String RESOURCES_BUNDLE = "resources.arsc";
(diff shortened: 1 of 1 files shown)


TombolaShepless:
@matrosov-nikita do you know when this will be merged and released? Also what is the status of the overall code push product? All the repos seem very low on activity with many issues stating "ran into issues so feature has been pushed back a few months". We want to choose Code Push for our company but this signs make us worried about whether this project is going to be around/supported in the long term?

silhouettes:
@TombolaShepless - apologies for the slow response! Just merged, and we hope to make a release sometime next week. As for your other concerns, I think they were addressed in the other issue, but feel free to comment if there's anything else you need :)

Issue #207, opened 2017-01-11
iOS 10.2: Old code inside www/ after appstore update


golovan:
After update from appstore, application still uses old codepush update which was downloaded earlier on device for previous version of application. It's the problem because our application will not work with BE until it (app) has the latest version. If we release codepush update for this new appstore version that helps but looks weird. As just after appstore update user sees application downloading update then codepush update process starts and user has to reload an app to make it work. We were able to repeat this issue on the latest iOS 10.2 (on all devices with this version). This's ionic1 cordova application. Is it possible to force new appstore release to use its own code and ignore previous codepush update? Do you have any thoughts when this might be managed/fixed?

sergey-akhalkov:
@egolovan - hi!
Agreed, releasing codepush update is weird solution, but for current version of `cordova-plugin-code-push` it is the only possible solution. There are two PRs should fix the issue, #201 and #202 are waiting to be merged and released. Hoping this will happen soon.

egolovan:
Thank you for your answer, @sergey-akhalkov! We tried #202 yesterday w/o any success. On iOS 10.2 that didn't work. ;( Hope there will be some solution soon!

TombolaShepless:
@sergey-akhalkov do you know when those PRs will be merged? I have put a comment on #201 expressing my concern regarding the activity of the CodePush project and its long term future. I don't mean to sound disrespectful/ungrateful - I just want to make a sounds decision for the business I work for.

sergey-akhalkov:
@TombolaShepless - hi!
Sorry for the silence on our side. Please see https://github.com/Microsoft/code-push/issues/366 answer relates to development state of CodePush solution. Unfortunately, for now I haven't any information about release plans, but will let you know if there is any news.

sergey-akhalkov:
@egolovan - hi!
I've tried to reproduce the issue with no luck with following repro steps:
1. Build and run app with version `1.0.0` having `"initial value"` string in `index.html` file.
2. Release and install an update - change `"initial value"` string to `old value` in `"index.html"` file.
3. Build and run app with version `1.0.1` having `"new value"` string in `index.html` file.

At the end I've got `index.html` with `"new value"` string. So could you please provide me with more info? (maybe `code-push-cli` version, `cordova-plugin-code-push` version - execute `code-push -v` and `cordova plugins ls` commands)

egolovan:
[...]

PR #211, opened 2017-02-03
Fixed issue for iOS when package folder was not cleared properly 

max-mironov:
This changes should fix https://github.com/Microsoft/cordova-plugin-code-push/issues/207. Tried to use more robust solution for determining correct build time for the app, Also add logic to also verify if the app version  was changed before clearing deployments. Also moved repeated code for getting the app version to utility function.


tatatananana:
Hi @max-mironov,

Please fix 3 calls from: [Utilities getAppVersion]  to [Utilities getApplicationVersion]
Ive tested it and It solves issue #207

max-mironov:
@tatatananana, thank you so much for pointing at this, changes are committed.

src/ios/Utilities.m
```diff
@@ -12,10 +16,18 @@ + (NSString*)getApplicationTimestamp{
     return nil;
 }
 
-+ (NSDate*)getApplicationBuildTime{
-    NSString *appPath = [[NSBundle mainBundle] bundlePath];
-    NSDictionary *executableAttributes = [[NSFileManager defaultManager] attributesOfItemAtPath:appPath error:nil];
-    return [executableAttributes objectForKey:@"NSFileModificationDate"];
++ (NSDate*)getApplicationBuildTime{
```
silhouettes: Is it possible that this change fixed the problem, without the need for the appVersion check above?

max-mironov:
@silhouettes I agree that it looks reasonable because in this way we will have some inconsistency in implementation this logic for iOS and Android (because we do not check for app version change on Android).

Check for appVersion is needed because the timestamp will not be updated unless we have some changes in native code, so if we bump app version in plist and rebuild than the timestamp won't be updated. I imagine this can be a problem because this way the app will not clear old deployments.

So you said that 
>Secondly, the app version can be different even if the binary has not been updated. This is because the version on the downloaded update can be a range, e.g. >= 1.0.0, whereas the version on the binary is always a fixed semver, e.g. 1.0.0.

I've double checked and it looks like app version is written to metadata here: https://github.com/Microsoft/cordova-plugin-code-push/blob/master/www/localPackage.ts#L177 (writeNewPackageMetadata function) and is acquired from native side by calling the same getAppVersion method so the app version couldn't be a range and check for appVersion will always make sense. Please correct me if I'm wrong.

silhouettes:
Thanks for the explanation @max-mironov, what you said makes sense. I'm sorry to be so nitpicky, but it seems like this fix doesn't fully cover the problem - if the dev makes a script-only change and deploys it to the simulator or device without bumping their app version, then it won't apply to them. Hopefully my understanding is correct here? It seems a bit much to ask people to bump their app version every time they make a script change to their app just so that they're able to test it.

It seems to me that the real issue here is, as you said, `the timestamp will not be updated unless we have some changes in native code`. Just brainstorming here, I'm wondering if we can copy what is done in the React Native plugin: https://github.com/Microsoft/react-native-code-push/blob/e42485243d9466fb6c00937bd9f67cee0d96de53/ios/CodePush/CodePush.m#L172? Not just for the sake of robustly addressing this issue, but also because it would be good to try and converge the behavior of the Cordova and React Native plugins as we will be attempting to unify the native codebase in the future.

Will that work? It may be that your approach is the only way to address the issue, but I just want to make sure we exhaust other options first.

silhouettes:
(Updated the above comment)

max-mironov:
@silhouettes thank you, I've got your point!
Digging deeper into this issue I've found out that we can use ios pre build script to touch file and this way timestamp will be updated on build every time despite we do not change any native code.
Shell script in PBXShellScriptBuildPhase section could be something like this:
`touch -m ${SRCROOT}/${PROJECT_NAME}/plugins/cordova-plugin-code-push/Utilities.m`

We can add Cordova "after_plugin_install" hook for ios in config.xml and it will execute script that will modify pbxproj file and insert BuildPhase run script section to it.

Please let me know what do you think about this solution?
If it works we can also apply the same solution for React Native iOS plugin.

silhouettes:
@max-mironov That's a good idea! Will that cause a re-compilation of the binary, and will it be a significant time cost? Also, would it work the same way if we just touched the binary after it's already been built?

silhouettes:
Okay, so I've been looking at this again, and I feel like there are two parts here that I didn't sort through clearly before:

1. App Version Check - My apologies, I do think it is good to include this now. This is partly for consistency with the RN plugin, which does the same check: https://github.com/Microsoft/react-native-code-push/blob/master/ios/CodePush/CodePush.m#L184. And partly just because this extra layer of safety is good for apps in production, as it absolutely guarantees they can never suffer this problem even if the timestamps are wrong (because the app version is always bumped in production).
2. Timestamp - I think it's important that this timestamp changes for both native and script changes, so that devs don't get weird behavior as they deploy to their simulator or device.

silhouettes:
Oh! I have an idea. What if we did all three of the following:

1. Check __DATE__ and __TIME__, which changes if native changes have been made
2. Check the `binaryHash` (hash of the JS that shipped with the binary) of the current binary, vs the `binaryHash` of the binary when the update was installed. This will change if JS changes have been made.
3. Check the appVersion (this is not necessary, but we include it as insurance)

In other words, I think that the `{ __DATE__, __TIME__, binaryHash }` tuple uniquely identifies a given binary. If the dev makes changes to either native or JS and then deploys their app, this tuple will have a different value. Do you think so?

max-mironov:
@silhouettes, ok finally tested the latest solution for getting correct build time by looking at NSModificationDate of plist file in main bundle. Verified it is working on iOS10.2.1 and iOS8.1 devices and is updated each time on native build and through cordova run ios deployment. Also tested this is working with TestFlight deployment. This should also cover the latest case you descried above.

PR #216, opened 2017-02-15
Set APK Build Time in Gradle

davidofwatkins:
This updates the mechanism that determines the APK build time to use a string resource from the Gradle build process instead of the date modified for classes.dex. It uses the same strategy that was implemented for react-native-code-push (see https://github.com/Microsoft/react-native-code-push/pull/683).

This is a second fix for #196 and handles the problem [discussed here](https://code.google.com/p/android/issues/detail?id=220039). It was partially fixed in a previous PR (#201), but that didn't fix it for me without `android.keepTimestampsInApk = true` in my gradle.properties. Either way, this seems like the more straightforward fix to me.


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

PR #224, opened 2017-02-28
Added app version check for Android for consistency with iOS approach

max-mironov:
This is for  'insurance'  that old packages will be cleared even if the timestamps are wrong because the app version is always bumped in production.

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -275,9 +275,19 @@ public class CodePush extends CordovaPlugin {
         if (deployedPackageMetadata != null) {
             String deployedPackageTimeStamp = deployedPackageMetadata.nativeBuildTime;
             long nativeBuildTime = Utilities.getApkBuildTime(this.cordova.getActivity());
-            if (nativeBuildTime != -1) {
+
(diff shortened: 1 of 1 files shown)


src/android/CodePush.java
```diff
@@ -275,9 +275,19 @@ private void clearDeploymentsIfBinaryUpdated() {
...
             String deployedPackageTimeStamp = deployedPackageMetadata.nativeBuildTime;
             long nativeBuildTime = Utilities.getApkBuildTime(this.cordova.getActivity());
-            if (nativeBuildTime != -1) {
+
+            String deployedPackageVersion = deployedPackageMetadata.appVersion;
+            String applicationVersion = null;
+            try {
+                applicationVersion = Utilities.getAppVersionName(this.cordova.getActivity());
+            } catch (PackageManager.NameNotFoundException e) {
+                e.printStackTrace();
```
silhouettes: I'm guessing we expect this exception should basically never happen right?

max-mironov: Yes, I thought about it too, I believe that if we don't have version defined than the app is not correct at all.

Issue #264, opened 2017-07-07
Android crash on startup after code-push update


Morathil:
### Description

Android crash after installing first code-push update.

### Reproduction

1.) Install app
2.) Release code push version
3.) Restart app (sync in background)
4.) Javascript console output:
```
[CodePush] Could not get application build time. Error: Could not get application timestamp.
[CodePush] Install succeeded.
[CodePush] Update is installed and will be run on the next app restart.
```
5.) Close app
6.) Open app
7.) Native crash 
```
Shutting down VM
FATAL EXCEPTION: main
Process: com.test.app, PID: 2996
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.test.app/com.test.app.MainActivity}: java
                                                            .lang.NumberFormatException: Invalid long: "1.4994191E12"
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2416)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2476)
at android.app.ActivityThread.-wrap11(ActivityThread.java)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1344)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:148)
at android.app.ActivityThread.main(ActivityThread.java:5417)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
Caused by: java.lang.NumberFormatException: Invalid long: "1.4994191E12"
at java.lang.Long.invalidLong(Long.java:124)
at java.lang.Long.parse(Long.java:363)
at java.lang.Long.parseLong(Long.java:353)
at java.lang.Long.parseLong(Long.java:321)
at com.microsoft.cordova.Utilities.getApkBuildTime(Utilities.java:62)
at com.microsoft.cordova.CodePush.clearDeploymentsIfBinaryUpdated(CodePush.java:277)
at com.microsoft.cordova.CodePush.onStart(CodePush.java:473)
at org.apache.cordova.PluginManager.onStart(PluginManager.java:273)
at org.apache.cordova.CordovaWebViewImpl.handleStart(CordovaWebViewImpl.java:458)
at org.apache.cordova.CordovaActivity.onStart(CordovaActivity.java:299)
[...]

PR #276, opened 2017-08-07
Fix incorrect long value restoration from strings.xml

ruslan-bikkinin:
Fix incorrect long value restoration from strings.xml due to android sdk bug
#264

### DIFF
--- a/src/android/Utilities.java
+++ b/src/android/Utilities.java
@@ -59,9 +59,11 @@ public class Utilities {
         Long millis;
 
         try {
+            //replace double quotes needed for correct restoration of long value from strings.xml
+            //https://github.com/Microsoft/cordova-plugin-code-push/issues/264
             millis = parseLong(context.getString(
                 context.getResources().getIdentifier("CODE_PUSH_APK_BUILD_TIME", "string", context.getPackageName())
-            ));
+            ).replaceAll("\"",""));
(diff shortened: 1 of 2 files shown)


src/android/Utilities.java
```diff
@@ -59,8 +59,10 @@ public static long getApkBuildTime(Context context) {
         Long millis;
 
         try {
+            //replace double quotes needed for correct restoration of long value from strings.xml
+            //https://github.com/Microsoft/cordova-plugin-code-push/issues/264
             millis = parseLong(context.getString(
-                context.getResources().getIdentifier("CODE_PUSH_APK_BUILD_TIME", "string", context.getPackageName())
+                context.getResources().getIdentifier("CODE_PUSH_APK_BUILD_TIME", "string", context.getPackageName().replaceAll("\"",""))
```
sergey-akhalkov: If I get it right, we should replace quotes for `context.getString(...)` method result, not for `context.getPackageName().replaceAll("\"","")`

sergey-akhalkov: If I get it right, we should replace quotes for context.getString(...) method result, not for context.getPackageName().replaceAll("\"","")

ruslan-bikkinin: yes, thanks for that

sergey-akhalkov:
lgtm, please feel free to merge.

PR #293, opened 2017-09-21
fix typescript compile error in remotePackage.ts

afoo:
After cloning the repository, then doing `npm install`, then `gulp`, the TS compilation failed with the error

```
www/remotePackage.ts(79,199): error TS2559: Type 'true' has no properties in common with type 'FileDownloadOptions'.
www/remotePackage.ts(79,199): error TS2559: Type 'true' has no properties in common with type 'FileDownloadOptions'.
```

Since the `noEmitOnError` TS compile option is set, this results in no JS files being emitted.

This commit fixes this by explicitly passing `null` for the `options` parameter, making the `true` be used for the `trustAllHosts` parameter, which is I believe what was intended.

### DIFF
--- a/bin/www/remotePackage.js
+++ b/bin/www/remotePackage.js
@@ -69,7 +69,7 @@ var RemotePackage = (function (_super) {
                         downloadProgress(dp);
                     }
                 };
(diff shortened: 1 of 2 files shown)


sergey-akhalkov:
Hi @afoo, the issue has place due to out of dated typings - I've fixed it in https://github.com/Microsoft/cordova-plugin-code-push/commit/ba61bc3a6b62cb4235dd46fe8f0fec2b2e02a950 - please verify and let me know if you see any issues.

PR #301, opened 2017-10-05
implement code signing for iOS, package hash checking for iOS and Android

afoo:
This patch implements checking of the `packageHash` of downloaded updates for Android and iOS and implements code signing for iOS.

PR #306, opened 2017-10-20
Fixed code signature verification for diff updates.

KamenGoranchev:
#### The Problem 
The current implementation was verifying the signature of the downloaded package that is located in the unzip folder. This was leading to wrong calculation of the hash of the incoming package in case just part of the files were coming with the update (diff update). The problem is related to #301.

#### The Solution
- The code signature verification was moved after the `handleDeployment` step in order to merge the upcoming files before the signature verification. That fixes the local hash calculation.
- The `.codepushrelease` file is no longer ignored during the coping of the new files, otherwise we end up with old signature file in the `deploy` folder. That is aligned with react native implementation where this file is not ignored as well. Note that the local hash calculation is still ignoring `.codepushrelease`  file during calculation, which is correct.


KamenGoranchev:
Can you please explain under what conditions we get an diff update and when full one?

ruslan-bikkinin:
Hi and Thanks for contribution @KamenGoranchev! The first update is always should be the full one, next updates will be diffed. In other words, diff job requires starting point from which diff should be computed.

bin/www/fileUtil.js
```diff
@@ -104,7 +104,7 @@ var FileUtil = (function () {
             var copyOne = function () {
                 if (i < entries.length) {
                     var nextEntry = entries[i++];
-                    if (nextEntry.name === ".DS_Store" || nextEntry.name === "__MACOSX" || nextEntry.name === ".codepushrelease") {
+                    if (nextEntry.name === ".DS_Store" || nextEntry.name === "__MACOSX") {
```
ruslan-bikkinin: I believe, we need to ignore `.codepushrelease` file too.

KamenGoranchev: I compared with `react-native-code-push` implementation and I don't see any files being ignored when unzip folder is merged to the deploy folder:
*IOS* - https://github.com/Microsoft/react-native-code-push/blob/e1cdd90e4ee6645a138eab8a43d383a37d50ef71/ios/CodePush/CodePushUpdateUtils.m#L111
 *Android* - https://github.com/Microsoft/react-native-code-push/blob/bb0e206149f2469426ca99259aa31848695add9a/android/app/src/main/java/com/microsoft/codepush/react/FileUtils.java#L18
Only case when this files are ignored is when the hash is calculated which we do as well.

Is there a reason why this files should be ignored only in the `cordova-plugin` implementation. If yes we need to make the code signing verification 2 step process:
1st - verify and compare the signature in the unzip folder against the one in the package metadata.
2nd - compute the local hash after the files are merged and compare it again against the expected hash.

ruslan-bikkinin: @KamenGoranchev i got it. I was confused that we are not ignoring JWT file during the hashing process, but you've moved it to copy process. I believe this is not a good idea to separate common logic of that comparing with react-native-code-push. We should not move ignore logic to copy process.

ruslan-bikkinin: @KamenGoranchev sorry, I was confused by the word "ignore".

> Is there a reason why this files should be ignored only in the cordova-plugin implementation. 
Yes, there was bug recently related with exception throwing during copying __MACOSX and .DS_Srore directories 
https://github.com/Microsoft/cordova-plugin-code-push/pull/215

I believe, your edit for selected line is correct.

PR #307, opened 2017-10-20
Code signing implemented for Android

KamenGoranchev:
Code signing implemented for Android. As a reference implementation was taken the react native implementation. The requirement is to have public key in the `config.xml` with a key `CodePushPublicKey`. That same as the ios implementation of code signing.

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -2,6 +2,10 @@ package com.microsoft.cordova;
 
 import android.content.pm.PackageManager;
 import android.os.AsyncTask;
+import android.util.Base64;
+
+import com.auth0.jwt.JWTVerifier;
+import com.auth0.jwt.JWTVerifyException;
 
 import org.apache.cordova.CallbackContext;
 import org.apache.cordova.ConfigXmlParser;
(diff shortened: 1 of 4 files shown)


ruslan-bikkinin:
Thank you for your contribution to CodePush Cordova SDK @KamenGoranchev! I'll be updating this branch during this PR reviewing and will provide you explanations of what I did.

src/android/CodePush.java
```diff
@@ -2,6 +2,10 @@
 
 import android.content.pm.PackageManager;
 import android.os.AsyncTask;
+import android.util.Base64;
+
+import com.auth0.jwt.JWTVerifier;
```
max-mironov: Probably we should use anither JWT library for decoding due to current library implementation consume a lot of methods. Please refer to this issue for details: https://github.com/Microsoft/react-native-code-push/issues/1002

src/android/CodePush.java
```diff
@@ -89,14 +102,93 @@ private boolean execVerifySignature(final CordovaArgs args, final CallbackContex
...
+                }
+
+                final PublicKey publicKey;
+                try {
+                    publicKey = parsePublicKey(stringPublicKey);
+                } catch (CodePushException e) {
+                    callbackContext.error("Error occured while creating the a public key" + e.getMessage());
+                    return null;
+                }
+
```
max-mironov: nit: probably can remove redundant blank line here and below

src/android/CodePushException.java
```diff
@@ -0,0 +1,26 @@
+package com.microsoft.cordova;
+
+/**
+ * Created by kgoranchev on 19/10/2017.
```
max-mironov: Please remove auto generated comments, don't use it throughout the project

src/android/CodePush.java
KamenGoranchev: Sorry about that. I was following the react native implementation, but for some reason I was looking at old code. If it is OK for you I will try with the same library they have now:
`com.nimbusds:nimbus-jose-jwt:5.1`

src/android/CodePushException.java
KamenGoranchev: Will do.

src/android/CodePush.java
max-mironov: That should be ok, thanks for staying in touch!

PR #310, opened 2017-10-24
Fixed binary hash calculation on android.

### LINKED ISSUE #303: Binary hash calculation fails on Android.
### Description
Binary hash calculation fails in Android native code in case there are sub-folder of the "assets/www" folder, which is more or less always for cordova application.

### Problem cause
The problem is happening because the check in the following line is now working correctly: 
https://github.com/Microsoft/cordova-plugin-code-push/blob/70b2ff80a51b1f40dd5e63c9b48be1522dcaa978/src/android/UpdateHashUtils.java#L69
`absolutePath.isDirectory()` is always returning `false` when the path is invalid, which is the case when we are computing the hash of the files in the asset folder. Having this check return `false`, also for directory will result in `IOException` later in the following line:
https://github.com/Microsoft/cordova-plugin-code-push/blob/70b2ff80a51b1f40dd5e63c9b48be1522dcaa978/src/android/UpdateHashUtils.java#L74

### Potential solutions
We need to find alternative solution to check whether an item in the asset folder is a file or directory, Unfortunately I couldn't find a clean solution so far. 
Workaround would be to try to open the file and catch `IOException`. For the cases we have exception we can assume that this was a Directory, but that is really ugly.

KamenGoranchev:
This PR is a proposal that fixes #303.

### The Solution
Since I couldn't find a easy and clean solution to determine if a the asset is a directory or not, I had a look how the `cordova-plugin-file` is solving this problem and according to the documentation recursion is not a good idea in the asset folder anyway: http://cordova.apache.org/docs/en/7.x/reference/cordova-plugin-file/index.html#slow-recursive-operations-for-android_asset.

My idea is to use the `cdvasset.manifest`, which we generate anyway during the build, to get a flat list containing the path to all the  asset within the `www` folder, which is later used for hash calculation.

The drawback is that we create a dependency to the `cdvCreateAssetManifest` build task, which should be clearly stated in case you accept my solution.

### DIFF
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -111,6 +111,8 @@ public class CodePush extends CordovaPlugin {
                         callbackContext.error("An error occurred when trying to get the hash of the binary contents. " + e.getMessage());
                     } catch (NoSuchAlgorithmException e) {
                         callbackContext.error("An error occurred when trying to get the hash of the binary contents. " + e.getMessage());
+                    } catch (ClassNotFoundException e) {
+                        callbackContext.error("An error occurred when trying to get the hash of the binary contents. " + e.getMessage());
                     }
 
                     return null;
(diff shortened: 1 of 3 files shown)


ruslan-bikkinin:
@KamenGoranchev thanks for your excellent contribution! LGTM.
btw
> The drawback is that we create a dependency to the cdvCreateAssetManifest build task, which should be clearly stated in case you accept my solution.

It is already a mandatory part of CodePush build gradle scenario, so no worries here.

https://github.com/Microsoft/cordova-plugin-code-push/blob/df50322281dfcc2763c0e46e92fe3e99f62d8dd3/src/android/build-extras.gradle#L22-L50

PR #313, opened 2017-11-02
Refactor code signing

ruslan-bikkinin:
This PR includes significant improvements and bug fixes to original code signing implementation authored by community contributors.


ruslan-bikkinin:
Hi @afoo @KamenGoranchev, we've tested code signing feature both for Android and iOS and it works just fine, great work, thank you guys!
We are going to release code-signing feature for Cordova, but we've faced with the following issue that we have to fix - current approach for adding JWT dependency for iOS SDK (via cocoa pods) imposes strict requirements:

1. users will have to install pod cli;
2. it increases minimal version of Cordova and cordova-ios for SDK.

It's very undesirable to introduce such kind of requirements/limitations so we're looking for workarounds (for example - add JWT as framework or static library). So if you guys have a chance to help us to resolve this issue or provide any ideas - we'll be able to release code-signing feature much faster.

afoo:
We could use the same approach as the react-native code-push plugin and just include the source of the JWT library with the plugin. That would put a slightly bigger burden of keeping it up to date with potentially security relevant fixes on you though.

src/android/Utilities.java
```diff
@@ -79,6 +79,10 @@ public static long getApkBuildTime(Context context) {
 
     public static void logException(Throwable e) {
         Log.e(CodePush.class.getName(), "An error occured. " + e.getMessage(), e);
+    }/media/rbikkinin/work/projects/cordova-code-signing/platforms/android/src/com/microsoft/cordova/CodePush.java
```
afoo: This looks like some kind of copy-paste mistake? Doesn't even compile.

sergey-akhalkov:
Hi @afoo, thanks for the advice, we've used this approach here https://github.com/Microsoft/cordova-plugin-code-push/pull/316

@afoo, @KamenGoranchev, just in case - we are going to release code-signing for cordova tomorrow or day after tomorrow.

sergey-akhalkov:
Hey guys @afoo, @KamenGoranchev - we've released code signing feature! Please verify
1. https://github.com/Microsoft/cordova-plugin-code-push/releases/tag/v1.10.0-beta
2. https://github.com/Microsoft/code-push/releases/tag/v2.1.2-beta
and let us know if you see any issues.

PR #335, opened 2017-12-21
Removed unnecessary dependency.

annakocheshkova:
Removed dependency that was causing the conflict during plugin installation (file dependency is already presented in file-transfer plugin).


ruslan-bikkinin:
LGTM

sergey-akhalkov:
@annakocheshkova, I think it would be better to use version 1.6.3 instead of 1.6.0 due to it contains some bugfixes that look important.

JFYI - in future we can remove `cordova-plugin-file-transfer` at all, please see https://cordova.apache.org/blog/2017/10/18/from-filetransfer-to-xhr2.html

PR #337, opened 2017-12-22
[Android] Update assets manifest file location for build-extras.gradle

ruslan-bikkinin:
This PR updates assets manifest file location in `build-extras.gradle` due to since cordova-android v7 its was changed

### DIFF
--- a/src/android/Utilities.java
+++ b/src/android/Utilities.java
@@ -23,7 +23,7 @@ import static java.lang.Long.parseLong;
  * Utilities class used for file and other common native operations.
  */
 public class Utilities {
-    private static final String ASSETS_MANIFEST_FILE = "cdvasset.manifest";
+    private static final String ASSETS_MANIFEST_FILENAME = "cdvasset.manifest";
 
     public static String readFileContents(File file) throws IOException {
         StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


src/android/Utilities.java
```diff
@@ -23,6 +23,8 @@
  * Utilities class used for file and other common native operations.
  */
 public class Utilities {
+    private static final String ASSETS_MANIFEST_FILE = "cdvasset.manifest";
```
sergey-akhalkov: `"cdvasset.manifest"` is actually filename, not the file itself, please use `ASSETS_MANIFEST_FILENAME` instead.

src/android/build-extras.gradle
```diff
@@ -20,7 +20,9 @@
        under the License.
  */
 ext.postBuildExtras = {
-    def inAssetsDir = file("assets")
+    
+    // Since cordova-android 7 assets directory moved to android/app/src/main/assets instead of android/assets
+    def inAssetsDir = file("src/main/assets").exists() ? file("src/main/assets") : file("assets")
```
sergey-akhalkov: I'd like to avoid of using code duplications, this could be changed like this:
```java
def inAssetsDir = file("src/main/assets");
if (!inAssetsDir.exists()) { // Add support for cordova-android < 7.0.0
    inAssetsDir = file("assets");
}
```

PR #339, opened 2017-12-25
Add stacktrace logging

ruslan-bikkinin:
This PR add stacktrace logging which is will be very handy to debug errors.

### DIFF
--- a/bin/www/codePushUtil.js
+++ b/bin/www/codePushUtil.js
@@ -35,11 +35,8 @@ var CodePushUtil = (function () {
         console.log(CodePushUtil.TAG + " " + msg);
     };
     CodePushUtil.logError = function (message, error) {
-        var errorMessage = message || "" + " " + CodePushUtil.getErrorMessage(error);
-        var stackTrace = '';
-        if (error.stack) {
-            stackTrace += '. Stacktrace: ' + error.stack;
-        }
+        var errorMessage = message || CodePushUtil.getErrorMessage(error);
+        var stackTrace = error.stack ? ". Stacktrace: " + error.stack : '';
         console.error(CodePushUtil.TAG + " " + errorMessage + stackTrace);
     };
     CodePushUtil.TAG = "[CodePush]";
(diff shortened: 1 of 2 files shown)


www/codePushUtil.ts
```diff
@@ -66,7 +66,11 @@ class CodePushUtil {
      */
     public static logError(message: String, error?: Error): void {
         var errorMessage = message || "" + " " + CodePushUtil.getErrorMessage(error);
-        console.error(CodePushUtil.TAG + " " + errorMessage);
+        var stackTrace = '';
```
sergey-akhalkov: It would be great to use `let` instead of `var` and `const` instead of `let` if possible.

www/codePushUtil.ts
```diff
@@ -66,7 +66,11 @@ class CodePushUtil {
...
+        if (error.stack) {
+            stackTrace += '. Stacktrace: ' + error.stack;
+        }
+        console.error(CodePushUtil.TAG + " " + errorMessage + stackTrace);
```
sergey-akhalkov: Let's use `template literals` for text formatting instead of `+`

bin/www/codePushUtil.js
```diff
@@ -36,7 +36,11 @@ var CodePushUtil = (function () {
     };
     CodePushUtil.logError = function (message, error) {
         var errorMessage = message || "" + " " + CodePushUtil.getErrorMessage(error);
-        console.error(CodePushUtil.TAG + " " + errorMessage);
+        var stackTrace = '';
+        if (error.stack) {
+            stackTrace += '. Stacktrace: ' + error.stack;
```
sergey-akhalkov: Stacktrace -> StackTrace

www/codePushUtil.ts
ruslan-bikkinin: Done, thx

www/codePushUtil.ts
ruslan-bikkinin: Done, thx

bin/www/codePushUtil.js
ruslan-bikkinin: Done, thx

PR #345, opened 2018-01-05
Added post install script.

annakocheshkova:
Added post install script to resolve different versions of file and file transfer plugin for different cordova versions.


scripts/beforeInstall.js
```diff
@@ -0,0 +1,20 @@
...
+    var deferral = new Q.defer();
+    
+    console.log("Detecting cordova version...");
+    var cordovaV = ctx.opts.cordova.version;
+
+    console.log("Cordova version is " + cordovaV);
+    console.log("Removing file plugin to ensure cordova compatibility...");
+    exec('cordova plugin remove cordova-plugin-file');
+    if (parseFloat(cordovaV) < 6.3) {
+        exec('cordova plugin add cordova-plugin-file-transfer@1.6.3');
```
ruslan-bikkinin: [exec](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) is async function, so we need to pass callback in order to handle exec result correctly.

sergey-akhalkov: @ruslan-bikkinin, good point! I would prefer to use sync version of this method - https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options

scripts/beforeInstall.js
```diff
@@ -0,0 +1,30 @@
...
+    var deferral = new Q.defer();
+
+    console.log("Detecting cordova version...");
+    var cordovaV = ctx.opts.cordova.version;    
+    console.log("Cordova version is " + cordovaV);
+
+    var plugins = exec('cordova plugin ls');
+    if (plugins.includes('cordova-plugin-file-transfer')) {
+        console.log("Removing file transfer plugin to ensure cordova compatibility... ");    
+        exec('cordova plugin remove cordova-plugin-file-transfer');
```
sergey-akhalkov: @annakocheshkova, I'm not sure if it's a good decision to remove user's installed plugins via script automatically, could you please investigate how Cordova resolve dependencies while installing plugins? I think we should provide only a logic that adds new plugins and avoid of logic that removes user's defined dependencies.

Issue #344, opened 2018-01-04
White screen appears and app breaks after second update from code-push [Ionic APP]


DLopezOpentix:
Hello,

### Description

I'm trying to use the plugin on an Ionic APP, following the steps indicated here (https://ionicframework.com/docs/native/code-push/).

Everything seems to work well, I have installed the plugin and it seems to be working. Then, I have released an update using the code-push CLI, and when I open the app again, I receive the update in the alert.

The first update seems to work well, the app restarts and code changes seem to have applied successfully.

However, when I update again, and send a new release, I receive the update in the app, but when the plugin tries to restart the app, a **white screen** appears and the **app breaks**. When I restart the app manually, **code changes are not applied**, and the update does not trigger again. If I check the updates in the devices with the code-push cli, it says **Rollbacks: 1**.

### Reproduction

`ionic start codepushtest blank`

`cd codepushtest`

`ionic cordova plugin add cordova-plugin-code-push`

`sudo npm install --save @ionic-native/code-push`

`ionic cordova platform add ios`

**This is how I release the update**:
`code-push release-cordova --description "Release 1" codepush-test ios`

In **src/app/app.module.ts**, in order to set up the CodePush service as a provider:

```typescript
import { BrowserModule } from '@angular/platform-browser';
import { ErrorHandler, NgModule } from '@angular/core';
import { IonicApp, IonicErrorHandler, IonicModule } from 'ionic-angular';
import { SplashScreen } from '@ionic-native/splash-screen';
import { StatusBar } from '@ionic-native/status-bar';

import { MyApp } from './app.component';
import { HomePage } from '../pages/home/home';

import { CodePush } from '@ionic-native/code-push';

@NgModule({
  declarations: [
    MyApp,
    HomePage
  ],
  imports: [
    BrowserModule,
    IonicModule.forRoot(MyApp)
  ],
  bootstrap: [IonicApp],
  entryComponents: [
    MyApp,
    HomePage
  ],
  providers: [
    StatusBar,
    SplashScreen,
    CodePush,
    {provide: ErrorHandler, useClass: IonicErrorHandler}
  ]
})
export class AppModule {}

```
In **src/app.component.ts**:

```typescript
import { Component } from '@angular/core';
import { Platform } from 'ionic-angular';
import { StatusBar } from '@ionic-native/status-bar';
import { SplashScreen } from '@ionic-native/splash-screen';

import { HomePage } from '../pages/home/home';

[...]

PR #346, opened 2018-01-08
Update docs with note about ionic apps specific

ruslan-bikkinin:
This PR adds note about how to release ionic apps.
Related to #344

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ code-push release-cordova MyApp-Android --privateKeyPath ~/rsa/private_key.pem
 
 The CodePush client supports differential updates, so even though you are releasing your app code on every update, your end users will only actually download the files they need. The service handles this automatically so that you can focus on creating awesome apps and we can worry about optimizing end user downloads.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +180,8 @@ code-push release-cordova MyApp-Android --privateKeyPath ~/rsa/private_key.pem
 
 The CodePush client supports differential updates, so even though you are releasing your app code on every update, your end users will only actually download the files they need. The service handles this automatically so that you can focus on creating awesome apps and we can worry about optimizing end user downloads.
 
+*NOTE: for **Ionic** apps you need to run `ionic build` before running `cordova-release` or `cordova` command in order to build web assets.*
```
sergey-akhalkov: @ruslan-bikkinin, could you please clarify why do we need to run `ionic build` command before the `cordova` command in terms of CodePush usage?

ruslan-bikkinin: @sergey-akhalkov this a typo, it should be `release` command instead. I'll fix it right now, thanks.

PR #361, opened 2018-02-05
Replace before-plugin-install hook to after-plugin-install

iTOYS:
Fix installation issue on Mac


svekl:
Tested on Windows, works good

PR #363, opened 2018-02-07
Added optional errback for codePush.sync()

### LINKED ISSUE #359: Codepush: Get failed sync error details
### Description

As of now, when the sync method is called and the SyncStatus get an ERROR value, there is no way to get the actual error. The documentation states the error details are in the console log but in live production, there should be a way to get the error details in the SyncCallback as an argument so they can be used for troubleshooting (like for eg. sending them out as error report to developers).

### Reproduction

Not relevant as this is an enhancement.

### Additional Information

* cordova-plugin-code-push version: All versions
* List of installed plugins: NA
* Cordova version: NA
* iOS/Android/Windows version: NA
* Does this reproduce on a debug build or release build? Both
* Does this reproduce on a simulator, or only on a physical device? Both

iTOYS:
Resolve #359


sergey-akhalkov:
Hey guys, from my point of view the CodePush API looks really weird now. Usually an `error` is the first arg for all callbacks.
I would prefer to leave the existing `sync` signature as it is and as workaround the customers can use the following approach: https://github.com/Microsoft/cordova-plugin-code-push/blob/master/samples/advanced/www/js/index.js

Please let me know what do you think.

iTOYS:
@sergey-akhalkov 
I'm absolutely agree. API looks really weird.
But the only possible way is a breaking changes, so I think we can merge this PR and going to change `sync` signature in next major release

sergey-akhalkov:
@iTOYS, I don't think that changing signatures from one release to another is a good approach and I believe that it would be better to release breaking changes right now because if I get it right you are going to release it anyway in the future.

As for breaking changes, we also can try to avoid them by using `typeof` to determine the type of the second argument:
```js
sync(syncCallback?: SuccessCallback<SyncStatus>, errorCallback?: ErrorCallback, syncOptions?: SyncOptions, downloadProgress?: SuccessCallback<DownloadProgress>): void;
```
And in case the type of the second arg is a "Function" - we are going to think that an user wants to use error callback and that the 3rd arg is about `syncOptions` and etc.
Otherwise if type of the second arg is an "Object" - user wants to specify `syncOptions` and etc.

Btw, I'm ok with any decision, just sharing my thoughts regarding this.

saifeer:
Hey @iTOYS 

It seems like the version in plugin.xml wasn't bumped up with this release so it still shows up as 1.11.3 instead of 1.11.4.

Thanks!

iTOYS:
Hi @saifeer,

Thanks for pointing this out.

iTOYS:
@saifeer 

1.11.5 was published

PR #408, opened 2018-05-08
Update file for backward compatibility

quintonn:
Updated hook to be backward compatible with older version of nodejs

### DIFF
--- a/hooks/afterPluginAdd.js
+++ b/hooks/afterPluginAdd.js
@@ -1,4 +1,5 @@
 module.exports = function (ctx) {
+
     if (ctx.cmdLine.includes("cordova-plugin-code-push")) {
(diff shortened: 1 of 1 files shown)


alexandergoncharov:
Hi @quintonn,

Thanks for contributing!

hooks/afterPluginAdd.js
```diff
@@ -1,4 +1,5 @@
 module.exports = function (ctx) {
+
     if (ctx.cmdLine.includes("cordova-plugin-code-push")) {
```
alexandergoncharov: Please replace this `includes()` method with `indexOf()`
Let's get rid of all `includes()` here for backword compatibility

quintonn:
Hi @alexandergoncharov,
I've replaced the other includes with an indexOf also.

quintonn:
@alexandergoncharov 
Done

PR #411, opened 2018-05-14
Fix issue with installing dependency in case if add plugin just in config.xml

alexandergoncharov:
* Prevent duplicated installing of dependency plugins.


alexandergoncharov:
Seems GitHub broke recognizing of changes.
Hope this screenshot of changes will be more understandable.
image

Issue #341, opened 2018-01-02
Drop dependency cordova-plugin-file-transfer


kaynz:
Hi!

### Description
The plugin cordova-plugin-file-transfer has recently been deprecated. Additionally it's causing errors if in the same project cordova-plugin-file ^6.0.0 is installed.

### Reproduction

Install cordova-plugin-file@latest
Install cordova-plugin-file-transfer@latest

### Additional Information

* cordova-plugin-code-push version: latest
* List of installed plugins: cordova-plugin-file:6.0.1
* Cordova version: 8.0.0
* iOS/Android version: 4.5.4/7.0.0

### Possible solutions

* Drop usage of cordova-plugin-file-transfer. Follow the migration guide and upgrade to XHR: https://cordova.apache.org/blog/2017/10/18/from-filetransfer-to-xhr2.html
* Fork cordova-plugin-file-transfer and raise version to ^6.0.0 (pretty good chance that new bugs will appear)
* Believe in god and hope that cordova merges https://github.com/apache/cordova-plugin-file-transfer/pull/195. Update to that new version in cordova-plugin-code-push.

sergey-akhalkov:
Hi @kaynz, thanks for reaching us! We've recently released [1.11.0](https://github.com/Microsoft/cordova-plugin-code-push/releases/tag/v1.11.0) version of Cordova Plugin for CodePush, so could you please update the plugin version in your dependencies and let us know if it resolves the issue?

fikriauliya:
I encounter this issue in 1.11.0

```
Discovered plugin "cordova-plugin-code-push" in config.xml. Adding it to the project
Installing "cordova-plugin-code-push" for ios
Installing "code-push" for ios
Installing "cordova-plugin-file" for ios
Installing "cordova-plugin-file-transfer" for ios
Failed to install 'cordova-plugin-file-transfer': CordovaError: Version of installed plugin: "cordova-plugin-file@6.0.1" does not satisfy dependency plugin requirement "cordova-plug
in-file@^5.0.0". Try --force to use installed plugin as dependency.
    at /Users/levi/.nvm/versions/node/v8.9.3/lib/node_modules/cordova/node_modules/cordova-lib/src/plugman/install.js:557:37
    at _fulfilled (/Users/levi/.nvm/versions/node/v8.9.3/lib/node_modules/cordova/node_modules/cordova-lib/node_modules/q/q.js:787:54)
[...]

PR #427, opened 2018-07-18
Add dependencies

libinvarghese:
Since there are certain dependencies used in the plugin and also installed via the hooks, to maintain consistency & clarity add  dependency in plugin.xml


alexandergoncharov:
Hi @libinvarghese ,
Thanks for contributing!

We tried to use this approach in this PR: https://github.com/Microsoft/cordova-plugin-code-push/pull/377 but we faced issue with versions if users would like to use plugin of other version https://github.com/Microsoft/cordova-plugin-code-push/issues/341.  And then we decided to put this all in hooks to make possible to update this plugin with other versions if user prefer.

Please let me know if you have any questions.

Thanks,
Alexander

PR #439, opened 2018-08-24
fixes #438

### LINKED ISSUE #438: Installation fails when local cordova is used
The hook script only checks for global cordova, and deferral is called when it hasn't been declared.

jpike88:


NickToropov:
Hi @jpike88,
Thanks for contributing!

Issue #486, opened 2018-12-22
Ionic Webview 2 with CodePush gives white screen on boot.


Silvest89:
Using the latest version(1.11.16) of CodePush I still get a white screen on boot. It seems it uses the file protocol and immediately gives me an empty index.html. Tested on IOS12 iPhone X Ionic 4
<img width="861" alt="screenshot 2018-12-22 at 03 18 55" src="https://user-images.githubusercontent.com/2388338/50369592-7190e580-0598-11e9-9f5b-6e5c929f2111.png">

And after using window.location.replace('http://localhost:8080/index.html') (remote debug device in safari.
Everything works until another restart.
<img width="928" alt="screenshot 2018-12-22 at 03 20 04" src="https://user-images.githubusercontent.com/2388338/50369601-99804900-0598-11e9-87c3-46a1247eafcf.png">

Everything works fine if I remove the cordova code push plugin

This is also easily reproducable in a start project.

ionic start codepush-ionic-test blank
<Pick Ionic4>
cd codepush-ionic-test
ionic cordova plugin add cordova-plugin-code-push
ionic cordova platform add ios
ionic cordova run ios or ionic cordova emulate ios

You will get a white screen on boot. Remove the plugin and everything works.

This happens with both Ionic Web View 2 and cordova-plugin-wkwebview-engine

### After further testing everything works fine if I use 
https://github.com/atftech-artyom/cordova-plugin-code-push

It seems that you removed 
``` Objective-C
  if ([self hasIonicWebViewEngine]) {
        [self setServerBasePath:url.path];
    } else {
        [self.webViewEngine loadRequest:[NSURLRequest requestWithURL:url]];
    }
```
In:
``` Objective-C
- (void)loadURL:(NSURL*)url {
    // In order to make use of the "modern" Cordova platform, while still
    // maintaining back-compat with Cordova iOS 3.9.0, we need to conditionally
    // use the WebViewEngine for performing navigations only if the host app
    // is running 4.0.0+, and fallback to directly using the WebView otherwise.
#ifdef __CORDOVA_4_0_0
    if ([self hasIonicWebViewEngine]) {
        [self setServerBasePath:url.path];
    } else {
        [self.webViewEngine loadRequest:[NSURLRequest requestWithURL:url]];
    }
#else
    [(UIWebView*)self.webView loadRequest:[NSURLRequest requestWithURL:url]];
#endif
```

How it is currently:
``` Objective-C
[...]

PR #491, opened 2019-01-17
Fix Ionic requests(white screen)

alexandergoncharov:
**Issue:** Ionic app should call all file URL requests with `setServerBasePath` method otherwise app shows just white screen.

**Solution:** Replaced calling `setServerBasePath` method to using it with all URL request.

Related issue: https://github.com/Microsoft/cordova-plugin-code-push/issues/486

### DIFF
--- a/src/ios/CDVWKWebViewEngine+CodePush.m
+++ b/src/ios/CDVWKWebViewEngine+CodePush.m
@@ -11,12 +11,12 @@
 
 - (id)loadRequest:(NSURLRequest *)request {
     NSURL *readAccessURL;
-    
+
     if (request.URL.isFileURL) {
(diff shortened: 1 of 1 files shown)


jp-uq:
Hi all,
since this merge the plugin doen´t work for ionic 3 anymore.
actual behaviour:
1. setServerBasePath is called for the new CodePush-Update URL
2. setServerBasePath is called for the ionic default URL

--> on every App Start the default App is displayed and the user never sees the updated version.

With the #483 PR all ionic Webview 2+ issues were solved, but for me now the plugin doesn´t work anymore.

Ionic webview version:
cordova-plugin-ionic-webview 2.3.1

Ionic info
bildschirmfoto 2019-01-30 um 08 58 28

PR #511, opened 2019-03-05
replace hardcoded index.html string with config defined content src

jtan743:
**Issue**: inside cordova config.xml if `<content src="index.html" />` is not index.html, after codepush update app will crash with StringIndexOutOfBoundsException

**Solution**: change hardcoded index.html to ConfigStartPageName


alexandergoncharov:
Hi @jtan743,
Thanks for contributing!

That's good catch! I'll test as soon as possible and let you know my results.

alexandergoncharov:
Hi @jtan743 ,

Sorry for this delay. I just tested and all works correctly. Thanks for contributing!

PR #526, opened 2019-04-29
Include "types" in the package.json

igorrmotta:
This change will enable the use of the types of this package.

``` 
/// <reference types="cordova-plugin-code-push" />
```


alexandergoncharov:
Hi @igorrmotta,
Thanks for contributing!

Good catch!

PR #584, opened 2019-11-15
Optimize memory management for package hashing

elijahr:
This patch prevents app memory from skyrocketing during sync if the local package is large.

My app began receiving a memory pressure warning event of 16, followed by a memory pressure warning event of 0, which would then result in the app being terminated. Memory usage quickly climbed to ~600mb after checking for CodePush updates. Profiling showed that most of the memory was being allocated underneath calls to `[UpdateHashUtils getHashForPath]`. I believe the primary culprit are the `fileContents` objects, which presumably are not being fully deallocated until after `[UpdateHashUtils addFolderEntriesToManifest]` completes its work. This would mean that the entire app package is being loaded into memory. This patch has stopped the memory pressure warnings, the app is not killed, and memory usage remains fairly constant around 16mb.


elijahr:
This is a rather small changeset that I'm sure would benefit other users of this plugin. I've met all CLA requirements. Please have a look and let me know if you need anything else from me or have any feedback.

alexandergoncharov:
Hi @elijahr ,
Thanks for contributing!

I tested your changes and found the issue by sending a request to the CodePush server. Looks like my test app can't send the request. I also tested this app with Master of cordova-plugin-code-push and successfully got the update. Here are my logs:
image

Have you faced this issue?
Please let me know if you have any questions.

elijahr:
@alexandergoncharov no I haven't run into this issue; we've been using this patch in production since November. Perhaps my changes need the latest from master, I'll pull.

All this code does is ensure that memory is garbage collected sooner.

elijahr:
@alexandergoncharov I've just pulled microsoft/master into this branch, perhaps try again?

alexandergoncharov:
@elijahr, yeah, it is working now. Thanks!

PR #599, opened 2020-02-13
Update the build-extras.gradle file

BrianMulhall:
Cordova Jira Ticket Exapling the Bug this fixes
https://issues.apache.org/jira/browse/CB-14163

We had trouble using this plugin on android with other plugins that also try to use ext.postBuildExtras.
We had cordova-support-google-services and cordova-plugin-code-push installed as well as this plugin and all three were trying to use postBuildExtras.
But only the last plugin listed in our config file would have its code executed at built time. the link above describes the cause of the issue pretty well

### DIFF
--- a/src/android/build-extras.gradle
+++ b/src/android/build-extras.gradle
@@ -19,7 +19,7 @@
        specific language governing permissions and limitations
        under the License.
  */
-ext.postBuildExtras = {
+cdvPluginPostBuildExtras.add({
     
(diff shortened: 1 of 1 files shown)


szh:
Any plans to merge this?

nonsenseless:
I'd like to emphasize the necessity of this update. Without it, you can't use phonegap-plugin-push in the same build. We've forked this plugin for our internal use, but it's a real issue for anybody that's not aware of it.

PR #608, opened 2020-04-15
fix(load): load different application path

videmort:
Currently, the plugin does not contemplate opening other index.html within a cordova application, only the main index.html

An example would be to open another html index from a plugin in cordova.  (swift)
`     
 let view = CDVViewController(nibName: nil, bundle: nil)
        view.wwwFolderName = "file://www"
        view.startPage = "mynewapp/index.html"
        view.modalPresentationStyle = .fullScreen
        let rootVIEW = UIApplication.shared.keyWindow?.rootViewController 
        rootVIEW?.presentedViewController?.dismiss(animated: false, completion: nil)
        rootVIEW?.present(view, animated: true, completion: nil)

`
So in that case your plugin will change the path to 
file://www/mynewapp/index.html and is NOT correct. because that file is not exist in that route. (plugin skip all the local server route, and for that is not working)

In the ionic weview plugin they calculate only one time  CDV_LOCAL_SERVER, but in this plugin every time that open one new application inside the base app.

 ̶*̶*̶T̶H̶I̶S̶ ̶P̶U̶L̶L̶ ̶I̶S̶ ̶N̶O̶T̶ ̶F̶O̶R̶ ̶M̶E̶R̶G̶E̶,̶ ̶T̶H̶I̶S̶ ̶S̶O̶L̶U̶T̶I̶O̶N̶ ̶W̶O̶R̶K̶S̶ ̶O̶N̶L̶Y̶ ̶F̶O̶R̶ ̶I̶N̶S̶T̶A̶L̶L̶ ̶A̶N̶D̶ ̶R̶E̶O̶L̶O̶A̶D̶ ̶T̶H̶E̶ ̶B̶A̶S̶E̶ ̶A̶P̶P̶ ̶B̶U̶T̶ ̶I̶S̶ ̶N̶O̶T̶ ̶A̶ ̶C̶L̶E̶A̶N̶ ̶S̶O̶L̶U̶T̶I̶O̶N̶,̶ ̶C̶H̶E̶C̶K̶I̶N̶G̶ ̶@̶"̶.̶a̶p̶p̶"̶ ̶a̶n̶d̶ ̶"̶@̶c̶o̶d̶e̶p̶u̶s̶h̶"̶ ̶T̶O̶ ̶C̶H̶E̶C̶K̶ ̶I̶F̶ ̶I̶S̶ ̶T̶H̶E̶ ̶B̶A̶S̶E̶ ̶A̶P̶P̶ ̶O̶R̶ ̶N̶O̶T̶ ̶I̶S̶ ̶N̶O̶T̶.̶*̶*̶ ̶ ̶
̶
̶*̶*̶T̶H̶I̶S̶ ̶S̶O̶L̶U̶T̶I̶O̶N̶ ̶F̶A̶I̶L̶S̶ ̶T̶O̶ ̶O̶P̶E̶N̶ ̶F̶I̶L̶E̶ ̶O̶N̶ ̶O̶T̶H̶E̶R̶'̶S̶ ̶P̶L̶U̶G̶I̶N̶S̶ ̶W̶H̶E̶N̶ ̶T̶H̶E̶ ̶P̶L̶U̶G̶I̶N̶ ̶I̶N̶S̶T̶A̶L̶L̶ ̶A̶ ̶N̶E̶W̶ ̶V̶E̶R̶S̶I̶O̶N̶*̶*̶

The function loadPluginRequest should be placed in the init step of this plugin and when plugin install new version should have same path not diferent one

**EDIT** 

Removing line --> navigateToLocalDeploymentIfExists in codepush.m file works fine now

### DIFF
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -341,6 +341,11 @@ StatusReport* rollbackStatusReport = nil;
 - (void)navigateToLocalDeploymentIfExists {
     CodePushPackageMetadata* deployedPackageMetadata = [CodePushPackageManager getCurrentPackageMetadata];
     if (deployedPackageMetadata && deployedPackageMetadata.localPath) {
+        NSString* startPage = ((CDVViewController *)self.viewController).startPage;
+        NSURL* URL = [self getStartPageURLForLocalPackage:deployedPackageMetadata.localPath];
+        if (![URL.path containsString:startPage]) {
+            return;
+        }
         [self redirectStartPageToURL: deployedPackageMetadata.localPath];
     }
 }
@@ -355,6 +360,7 @@ StatusReport* rollbackStatusReport = nil;
         [self handleUnconfirmedInstall:NO];
     }
 
+    [self navigateToLocalDeploymentIfExists];
     // handle both ON_NEXT_RESUME and ON_NEXT_RESTART - the application might have been killed after transitioning to the background
     if (pendingInstall && (pendingInstall.installMode == ON_NEXT_RESTART || pendingInstall.installMode == ON_NEXT_RESUME)) {
         [self markUpdate];
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -355,7 +355,6 @@ StatusReport* rollbackStatusReport = nil;
         [self handleUnconfirmedInstall:NO];
     }
 
-    //[self navigateToLocalDeploymentIfExists];
     // handle both ON_NEXT_RESUME and ON_NEXT_RESTART - the application might have been killed after transitioning to the background
     if (pendingInstall && (pendingInstall.installMode == ON_NEXT_RESTART || pendingInstall.installMode == ON_NEXT_RESUME)) {
         [self markUpdate];
--- a/src/ios/CodePush.m
+++ b/src/ios/CodePush.m
@@ -355,7 +355,7 @@ StatusReport* rollbackStatusReport = nil;
         [self handleUnconfirmedInstall:NO];
     }
 
-    [self navigateToLocalDeploymentIfExists];
+    //[self navigateToLocalDeploymentIfExists];
     // handle both ON_NEXT_RESUME and ON_NEXT_RESTART - the application might have been killed after transitioning to the background
     if (pendingInstall && (pendingInstall.installMode == ON_NEXT_RESTART || pendingInstall.installMode == ON_NEXT_RESUME)) {
         [self markUpdate];
--- a/src/android/CodePush.java
+++ b/src/android/CodePush.java
@@ -420,7 +420,7 @@ public class CodePush extends CordovaPlugin {
         CodePushPackageMetadata deployedPackageMetadata = this.codePushPackageManager.getCurrentPackageMetadata();
         if (deployedPackageMetadata != null && deployedPackageMetadata.localPath != null) {
             File startPage = this.getStartPageForPackage(deployedPackageMetadata.localPath);
-            if (startPage != null) {
+            if (startPage != null && this.isMainBundleActivity()) {
                 /* file exists */
                 try {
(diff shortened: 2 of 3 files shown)


GusLAN:
+1

pepedeka:
+1

videmort:
> Currently, the plugin does not contemplate opening other index.html within a cordova application, only the main index.html
> 
> An example would be to open another html index from a plugin in cordova. (swift)
> `
> let view = CDVViewController(nibName: nil, bundle: nil)
> view.wwwFolderName = "file://www"
> view.startPage = "mynewapp/index.html"
> view.modalPresentationStyle = .fullScreen
> let rootVIEW = UIApplication.shared.keyWindow?.rootViewController
> rootVIEW?.presentedViewController?.dismiss(animated: false, completion: nil)
> rootVIEW?.present(view, animated: true, completion: nil)
> 
> `
> So in that case your plugin will change the path to
> file://www/mynewapp/index.html and is NOT correct. because that file is not exist in that route. (plugin skip all the local server route, and for that is not working)
> 
> In the ionic weview plugin they calculate only one time CDV_LOCAL_SERVER, but in this plugin every time that open one new application inside the base app.
> 
> **THIS PULL IS NOT FOR MERGE, THIS SOLUTION WORKS ONLY FOR INSTALL AND REOLOAD THE BASE APP BUT IS NOT A CLEAN SOLUTION, CHECKING @".app" and "@codepush" TO CHECK IF IS THE BASE APP OR NOT IS NOT.**
> 
> **THIS SOLUTION FAILS TO OPEN FILE ON OTHER'S PLUGINS WHEN THE PLUGIN INSTALL A NEW VERSION**
> 
> The function loadPluginRequest should be placed in the init step of this plugin and when plugin install new version should have same path not diferent one
> 
> **EDIT**
> 
> Removing line --> navigateToLocalDeploymentIfExists in codepush.m file works fine now

Removed  check @".app" and "@codepush"  for --> bundleURL .

Works fine and can be merged

arturoberanlg:
Agree +1

src/ios/CodePush.m
```diff
@@ -355,7 +355,7 @@ - (void)pluginInitialize {
         [self handleUnconfirmedInstall:NO];
     }
 
-    [self navigateToLocalDeploymentIfExists];
+    //[self navigateToLocalDeploymentIfExists];
```
matthiaswenz, line 358: If this is not needed anymore, we can probably remove it altogether. If there might still be cases where we need to call this, we might need some more changes here.

videmort: Removed.
I don't think it is necessary to call that method. I tried all the possibilities and that method do nothing

Krasavinigor:
Hello @videmort!
Sorry for delay. Thank you for contributing. You've done good and useful job :smiley:

I tested your PR and checked that installation with ON_NEXT_RESTART mode works incorrectly, bcs app uses the old bundle. I have logs where all steps were done successfully.
All is good with IMMEDIATE mode. 
Could you please investigate this issue??

videmort:
@Krasavinigor hello again. I see something more. In INMEDIATE mode, first time Works perfect as you saw it, but if close and open the application it load the "base" version of the app. 
I'm checking but I don't understand very well the code of this plugin.

Can you and your team check it?

Krasavinigor:
@videmort 
I've got this error. I think it's the same issue as in my case. App uses the old bundle after a restart.

videmort:
@Krasavinigor  I don't see the point that the plugin do that

videmort:
@Krasavinigor it's fixed now

videmort:
@Krasavinigor  @alexandergoncharov  new release?

alexhisen:
Note that this change breaks build with cordova-android older than 6.4 due to the use of getContext() call.

PR #620, opened 2020-06-02
Fix H6 tag in readme

matt-auckland:
There were too many # so it wasn't rendering an h6

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,7 +602,7 @@ The optional `downloadProgress` callback is invoked several times during the dow
 
     Defines the format of the DownloadProgress object, used to send periodical update notifications on the progress of the update download.
 
-    ####### Properties
+    ###### Properties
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -602,7 +602,7 @@ The optional `downloadProgress` callback is invoked several times during the dow
 
     Defines the format of the DownloadProgress object, used to send periodical update notifications on the progress of the update download.
 
-    ####### Properties
+    ###### Properties
```
ruslan-bikkinin, line 605:
```suggestion
    Properties
```

ruslan-bikkinin:
Hi @matt-auckland, I've made a small correction and removed H6 tag at all to prevent it to be rendered on the same section level as `DownloadProgress`. Anyway, thanks a lot for pointing out to this and making docs better 👍

ruslan-bikkinin (approved):

PR #631, opened 2020-07-20
Prevent error when publicKey or signature is an empty string

darron1217:
I was experiencing the error below
```
Error! Public key was provided...
```

But I've never put public/private key during deploy&update.

I figured out that Capacitor(successor of Cordova) returns empty string on public key request.
Empty string also means not providing public key, so I changed the code to cover that case

### DIFF
--- a/bin/www/localPackage.js
+++ b/bin/www/localPackage.js
@@ -140,13 +140,13 @@ var LocalPackage = (function (_super) {
                 return;
             }
             publicKey = publicKeyResult;
-            isSignatureVerificationEnabled = (publicKey !== null);
+            isSignatureVerificationEnabled = !!publicKey;
             _this.getSignatureFromUpdate(deploymentResult.deployDir, function (error, signature) {
                 if (error) {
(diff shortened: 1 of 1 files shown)


alexandergoncharov:
Hi @darron1217 ,
Thanks for contributing!

It is a good catch!

bin/www/localPackage.js
```diff
@@ -140,13 +140,13 @@ var LocalPackage = (function (_super) {
                 return;
             }
             publicKey = publicKeyResult;
-            isSignatureVerificationEnabled = (publicKey !== null);
+            isSignatureVerificationEnabled = !!publicKey;
```
alexandergoncharov, line 143: You made changes in `bin` folder. This folder is generated by compiling code. Could you please make changes in source code(`www` folder) and run `gulp compile` command?

darron1217: Got it.

darron1217:
I created new PR cuz master has changed

PR #635, opened 2020-07-28
Compatibility with cordova-ios@6.1

szh:
Fixes #611 and the final remaining issue in #624 

I'm not experienced with Objective-C so please review my code carefully!

Issue #624, opened 2020-06-03
Not compatible with cordova-ios 6


dpogue:
### Description

The cordova-plugin-code-push plugin depends on the [deprecated (since 2017)](https://cordova.apache.org/blog/2017/10/18/from-filetransfer-to-xhr2.html) cordova-plugin-file-transfer plugin, which [is not compatible](https://github.com/apache/cordova-plugin-file-transfer/issues/258) with cordova-ios 6 and results in compiler issues.

The ideal solution here would be for cordova-plugin-code-push to handle the downloading, unzipping, and installation of app bundles in native code, that same way that [react-native-code-push](https://github.com/microsoft/react-native-code-push) does.

### Reproduction

Install cordova-plugin-code-push in an app built with cordova-ios 6.
There are compiler errors in the cordova-plugin-file-transfer Objective C files.

### Additional Information

* cordova-plugin-code-push version: 1.12.4
* List of installed plugins: cordova-plugin-code-push
* Cordova version: CLI 9.0.0 with cordova-ios@6.0.0
* iOS/Android/Windows version: iOS 13.5
* Does this reproduce on a debug build or release build? Both
* Does this reproduce on a simulator, or only on a physical device? Both

davidellolio:
Hi,

I have the same error.

D.

juanlurie:
I’m having the same problem. Juggling plugin versions for IOS with the WebKit restriction has led me to remove the cordova-plugin-file-transfer but that breaks code push

sectionDayongtek:
Hi, I have the same problem. 

help plz...

juanlurie:
I downgraded cordova-ios to 5.1.1 and added file transfer back and everything seems to be working now. The eco system is so fragile.

jacobg:
Related to these issues #485 and #341. There's already a pull request #513 to remove file transfer, but Microsoft has neglected to merge it.

swbradsh:
cordova-ios 6 also removed CDVUserAgentUtil, which this plugin is still referencing.

bobrosoft:
`Use of undeclared identifier 'CDVUserAgentUtil'`
Switched to cordova-ios 5.1.1 because of that issue 🤦  https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/561

videmort:
[...]

PR #655, opened 2020-10-07
Fix cordova-ios 6 issue with file:// requests being rejected

ruslan-bikkinin:
Related issue #624.

This PR is similar to #635 but makes a fix more clear way.

## Problem

Starting with cordova-ios 6 [`UIWebView` is replaced with `WKWebView` and also `WKURLSchemeHandler` is introduced](https://cordova.apache.org/announcements/2020/06/01/cordova-ios-release-6.0.0.html). 
If add these parameters to `config.xml`
```xml
<preference name="scheme" value="app" />
<preference name="hostname" value="localhost" />
```
then `file://` requests are rejected by `WKURLSchemeHandler` which causes app to become blank after applying CodePush update.

There was already [a fix](https://github.com/microsoft/cordova-plugin-code-push/blob/master/src/ios/CDVWKWebViewEngine%2BCodePush.m#L56-L84) for that in `CDVWKWebViewEngine+CodePush.m` but it was applied only for the case when 
* cordova-ios 4, 5 was used
* Additional [cordova-plugin-wkwebview-engine](https://github.com/apache/cordova-plugin-wkwebview-engine) package was installed and configured within application (it was supposed to be used as a temporary solution for cordova-ios 4, 5 to let developers use `WKWebView` unless official support is added)

This PR applies the fix for cordova-ios 6 as well.

## Implementation specifics

* `loadRequest` and `loadPluginRequest ` methods was extracted from `CDVWKWebViewEngine+CodePush.m` into new `WebViewShared` class. Now both `CodePush.m` and `CDVWKWebViewEngine+CodePush.m` shares the same fix logic written in one place.
* This class works as a [thread-safe singleton](https://developer.apple.com/documentation/dispatch/1447169-dispatch_once) to avoid it's possible re-initialization during the CodePush plugin life cycle.
* To avoid showing blank screen after applying the update and restarting application a workaround was implemented (please see comment [here](https://github.com/microsoft/cordova-plugin-code-push/pull/655/files#diff-bd9033e830cd10b9ac2ba5cec7f741b1R23-R41) for the explanation)

### DIFF
--- a/src/ios/CDVViewController+CodePush.m
+++ b/src/ios/CDVViewController+CodePush.m
@@ -40,7 +40,7 @@
      Note 1: self.webViewEngine loadRequest will be failing anyway but calling [webViewShared loadRequest] will hide it for user.
      Note 2: It affects only cordova-ios 6 apps
     */
-    if([Utilities cordova6OrGreater]) {
+    if([Utilities CDVWebViewEngineAvailable]) {
         WebViewShared* webViewShared = [WebViewShared getInstanceOrCreate:self.webViewEngine
                                                   andCommandDelegate:self.commandDelegate
                                                    andViewController:self];
--- a/src/ios/CDVViewController+CodePush.m
+++ b/src/ios/CDVViewController+CodePush.m
@@ -1,0 +1,81 @@
+#import <Cordova/CDVViewController.h>
+#import <objc/runtime.h>
+#import "WebViewShared.h"
+#import "Utilities.h"
+
+@interface CDVViewController (CodePush)
+- (NSURL*)appUrl; // Expose private method from original CDVViewController
+@end
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wincomplete-implementation"
+
+@implementation CDVViewController (CodePush)
+
+#pragma clang diagnostic pop
+
+- (void)viewDidLoad_codepush
+{
+    // Calls original viewDidLoad method from CDVViewController
+    [self viewDidLoad_codepush];
+    
+    /*
+     In original viewDid method from CDVViewController at the end it tries to open using something like self.webViewEngine loadRequest [self appUrl].
+     [self appUrl] is only resolves full path for startPage.
+     For example
+     
+     // CDVViewController.m
+     ...
+     // self.startPage is set somewhere to @"index.html";
+     
+     - (void)viewDidLoad {
+        ...
+         NSURL* url = [self appUrl]; //manipulates with self.startPage
+         //url is now @"file://HOSTNAME/path/to/index.html"
+     }
+     
+     It leads to WKWebView unable to load it because url prefixed with "file://" is rejected by web view for some reason.
+     In practice it means that after applying CodePush update and restarting app blank screen is appeared.
+     To workaround it [webViewShared loadRequest] is called after self.webViewEngine loadRequest is called in the original viewDidLoad method.
+     Note 1: self.webViewEngine loadRequest will be failing anyway but calling [webViewShared loadRequest] will hide it for user.
+     Note 2: It affects only cordova-ios 6 apps
+    */
+    if([Utilities cordova6OrGreater]) {
+        WebViewShared* webViewShared = [WebViewShared getInstanceOrCreate:self.webViewEngine
+                                                  andCommandDelegate:self.commandDelegate
+                                                   andViewController:self];
+        NSURL* url = [self appUrl];
+        [webViewShared loadRequest:[NSURLRequest requestWithURL:url]];
+    }
+}
+
+// Swizzling original viewDidLoad method
++ (void)load
+{
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        Class class = [self class];
+
+        SEL originalSelector = @selector(viewDidLoad);
+        SEL swizzledSelector = @selector(viewDidLoad_codepush);
+
+        Method originalMethod = class_getInstanceMethod(class, originalSelector);
+        Method swizzledMethod = class_getInstanceMethod(class, swizzledSelector);
+
+        BOOL didAddMethod = class_addMethod(class,
+                                            originalSelector,
+                                            method_getImplementation(swizzledMethod),
+                                            method_getTypeEncoding(swizzledMethod));
+
+        if (didAddMethod) {
+            class_replaceMethod(class,
+                                swizzledSelector,
+                                method_getImplementation(originalMethod),
+                                method_getTypeEncoding(originalMethod));
+        } else {
+            method_exchangeImplementations(originalMethod, swizzledMethod);
+        }
+    });
+}
+
+@end
--- a/src/ios/CDVViewController+CodePush.m
+++ b/src/ios/CDVViewController+CodePush.m
@@ -20,7 +20,7 @@
     [self viewDidLoad_codepush];
     
     /*
-     In original viewDid method from CDVViewController at the end it tries to open using something like self.webViewEngine loadRequest [self appUrl].
+     In original viewDid method from CDVViewController at the end it tries to open start page using something like self.webViewEngine loadRequest [self appUrl].
      [self appUrl] is only resolves full path for startPage.
      For example
      
--- a/src/ios/Utilities.m
+++ b/src/ios/Utilities.m
@@ -24,6 +24,10 @@
     return fileDate;
 }
 
++ (BOOL)cordova6OrGreater{
+    return NSClassFromString(@"CDVWebViewEngine");
+}
+
 void CPLog(NSString *formatString, ...) {
     va_list args;
     va_start(args, formatString);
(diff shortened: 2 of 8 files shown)


alexandergoncharov (approved):

sithwarrior:
A couple of comments, as stated in the issues, **UIWebView**, is not only deprecated, its persona non grata.

So the backwards comp in CodePush.m
```
        // In order to make use of the "modern" Cordova platform, while still
        // maintaining back-compat with Cordova iOS 3.9.0, we need to conditionally
        // use the WebViewEngine for performing navigations only if the host app
        // is running 4.0.0+, and fallback to directly using the WebView otherwise.
        #if (WK_WEB_VIEW_ONLY && defined(__CORDOVA_4_0_0)) || defined(__CORDOVA_4_0_0)
            [self.webViewEngine loadRequest:[NSURLRequest requestWithURL:url]];
        #else
            [(UIWebView*)self.webView loadRequest:[NSURLRequest requestWithURL:url]];
        #endif
```

that mentions **UIWebView**, is actualy a showstopper. New Apps with this code will be rejected by the App store.
App updates with this code, will be rejected by December.

Doing a search, shows thats the CodePushReportingManager.m also has **UIWebView** code

Both should be easy, to just cut?
No one needs backwards comp with old Cordova, if you need apps on the App store.
However, as this is a breaking change, Major version change is maybe warranted?

Here is the post from Apple, about not having UIWebView code: [https://developer.apple.com/news/?id=12232019b](url)

ruslan-bikkinin:
Hi @sithwarrior yeah, we are aware about that. The changes in this PR are only to fix broken cordova 6 issues for customers. Removing of `UIWebView` is in our TODO list for the next release (after the upcoming `v1.13.2`).

Wouter33:
Hi @ruslan-bikkinin, thanks for this amazing PR. So you are aware that this PR does not allow us to submit new apps right? Only update existing apps for now. 

In my opinion, priority should be to support new (and by definition also existing) cordova-ios 6 apps with a new major version which removes all UIWebView mentions. People using cordova-ios 5 and below can still use older versions of this library. Unless this is significant more work for now and you just want to make the PR available for existing apps.

andreidubov (approved):

src/ios/Utilities.m
```diff
@@ -24,6 +24,10 @@ + (NSDate*)getApplicationBuildTime{
     return fileDate;
 }
 
++ (BOOL)cordova6OrGreater{
+    return NSClassFromString(@"CDVWebViewEngine");
```
snowpardx, line 28: The function name is a bit confusing, because we actually check not for the `cordova 6`, but for existence of `CDVWebViewEngine` class, so if it get retired in e.g. cordova 7 it would be hard to identify just from code.

Can we please rename to something like `CDVWebViewEngineSupported` and cache the return value, so that subsequent calls were cheap?

snowpardx (approved):

src/ios/CDVViewController+CodePush.m
```diff
@@ -0,0 +1,81 @@
...
+
+- (void)viewDidLoad_codepush
+{
+    // Calls original viewDidLoad method from CDVViewController
+    [self viewDidLoad_codepush];
+    
+    /*
+     In original viewDid method from CDVViewController at the end it tries to open start page using something like self.webViewEngine loadRequest [self appUrl].
+     [self appUrl] is only resolves full path for startPage.
+     For example
```
Krasavinigor, line 25:
```suggestion
     For example:
```

Krasavinigor (approved):
Looks good to me 😄

src/ios/Utilities.m
ruslan-bikkinin: Thanks, done.

ruslan-bikkinin:
Hi @Wouter33 good point. I'm going to start working on removing `UIWebView` remnants today. There will be one new version `2.0.0` that accumulates all cordova-ios changes.

sithwarrior:
@ruslan-bikkinin  I just tried installing this directly from the branch

`https://github.com/microsoft/cordova-plugin-code-push#cordova-ios6-fix`

It dosent work for me, it goes straight to the blank page.

I've tried my poor objective C debugging skills, and it seems there is no good case for a Cordov-IOS 6 without the ionic WKWebView? (Cordova ios 6 has builtin WKwebview)

The app is now loaded from file:// instead of the custom url scheme.
This causes a bunch of CORS errors when the app loads, and breaks the app.

This is actually even worse than before.

ruslan-bikkinin:
@sithwarrior what's Cordova/Ionic version you are using? Could you please also share your `config.xml` and `package.json` (you can strip all sensitive data)?

ruslan-bikkinin:
@sithwarrior If you also could attach demo app on which you are experiencing the issue that would be very helpful!

sithwarrior:
> @sithwarrior If you also could attach demo app on which you are experiencing the issue that would be very helpful!

I just reproduced this on the codepush basic app, included with this project.

Steps to reproduce:
use Cordova-iOS 6.1

Setup custom URL scheme as recommended in config.xml on iOS
`<preference name="scheme" value="app" />
        <preference name="hostname" value="localhost" />`

remove the codepush plugin
`cordova plugin rm cordova-plugin-code-push`

Install codepush from master, as the PR is now merged
`cordova plugin add https://github.com/microsoft/cordova-plugin-code-push`

Build and run.

The app is now served from file://
This is a big no no on ios, as CORS on iOS blocks everything served from file://

ruslan-bikkinin:
@sithwarrior I just tried to reproduce issue using the steps you provided and faced no problems:
image

May be iOS version is involved. Do you use simulator or real device? What iOS version of device/simulator did you use?

sithwarrior:
@ruslan-bikkinin Yes the app loads, but if you use Safari remote debugger, you can see that with the new changes in the PR, the app is now served from file://

That should not happen, as CORS will not work, in any way.
It should be hosted from the custom url scheme, defined in the config.xml ie app://localhost

ruslan-bikkinin:
@sithwarrior could you please clarify what should I do in the basic app to start running into CORS issue you're describing?

sithwarrior:
@ruslan-bikkinin 
The core of the problem is that the custom url scheme is not beeing used in this update. (it is in the current released version)

hosting from file:// creates a lot of problems on ios, thats why the ionic wkwebview uses a built-in http server or custom url schemes, and thats why the new **cordova-ios 6**, also uses this feature.

`Additionaly, WKURLSchemeHandler support has been introduced with this release. Using a custom scheme to serve your app content through fixes CORS issues that exist because of the strict security policies that WKWebView has applied to the file scheme. You can easily configure your Cordova project to use a custom scheme by setting the preference options scheme and hostname in the config.xml file.`
[https://cordova.apache.org/announcements/2020/06/01/cordova-ios-release-6.0.0.html](url)


The problem is most apparent, when you need to load additional files, such as json config files, local media etc.
We in our production app, use angular translate, with .json files for the translations, and multiple .html files for our templates.

serving the app from file:// results in this:
Screenshot 2020-10-12 at 18 03 07

sithwarrior:
Im not a an objective C developer, so I dont no if it helps, but you can see here, how cordova-ios 6 loads the url scheme

[https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Private/Plugins/CDVWebViewEngine/CDVWebViewEngine.m](url)

```
    // viewController would be available now. we attempt to set all possible delegates to it, by default
    CDVViewController* vc = (CDVViewController*)self.viewController;
    NSDictionary* settings = self.commandDelegate.settings;

    NSString *scheme = [settings cordovaSettingForKey:@"scheme"];

    // If scheme is file or nil, then default to file scheme
    self.cdvIsFileScheme = [scheme isEqualToString: @"file"] || scheme == nil;

    NSString *hostname = @"";
    if(!self.cdvIsFileScheme) {
        if(scheme == nil || [WKWebView handlesURLScheme:scheme]){
            scheme = @"app";
        }
        vc.appScheme = scheme;

        hostname = [settings cordovaSettingForKey:@"hostname"];
        if(hostname == nil){
            hostname = @"localhost";
        }

        self.CDV_ASSETS_URL = [NSString stringWithFormat:@"%@://%@", scheme, hostname];
    }
```

ruslan-bikkinin:
Thanks a lot @sithwarrior, now I see the problem. I'm going to revisit this and fix as well.

ruslan-bikkinin:
Well, the thing is that all app assets eventually bundled into app binary (that's why if you are using custom scheme then WebView can resolve them), but Code Push can't obviously alter app binary and have to store bundle somewhere on the file system. It's not clear for us yet how to configure Web View to force it to allow refer to local Code Push bundle via `app://localhost`. I guess Ionic couldn't do something about it as well so they came up with custom local server that do the trick. I'm currently in search of a solution for that rather than introducing local server for CodePush assets. In the meantime if anybody from the community have any ideas how to workaround it, please let us know.

PR #658, opened 2020-10-15
Fix cordova-ios v6 issues and remove UIWebView support

ruslan-bikkinin:
Related issue #624.

What this PR does:

* Starting with cordova-ios 6 [`UIWebView` is replaced with `WKWebView` and also `WKURLSchemeHandler` is introduced](https://cordova.apache.org/announcements/2020/06/01/cordova-ios-release-6.0.0.html) which allow users to specify app scheme to avoid `file://` requests (that causes CORS issues). Previously Code Push update was loaded on WebView using `file://` requests, but now if app scheme is configured plugin will be using it to load the updated app page.

* Apple completely [drops support for UIWebView as of December 2020](https://developer.apple.com/news/?id=12232019b). PR removes support for UIWebView as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,11 @@ In order to ensure that your end users always have a functioning version of your
 Cordova 5.0.0+ is fully supported, along with the following associated platforms:
 
 * Android ([cordova-android](https://github.com/apache/cordova-android) 4.0.0+) - *Including CrossWalk!* *Note: Only on TLS 1.2 compatible devices*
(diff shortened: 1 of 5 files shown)


ruslan-bikkinin:
Hi @sithwarrior if you have time could you please check this on your app?

snowpardx (approved):

Krasavinigor (approved):

sithwarrior:
@ruslan-bikkinin after some testing. it seems to be working nicely, the app is hosted correctly, and local files and media, loads as expected, updates also work as expected. 

Im going to test it some more tomorrow.

andreidubov (approved):

alexandergoncharov (approved):

alexChernyatiev (approved):

Issue #679, opened 2021-06-24
cordova-plugin-code-push dependency overrides AFNetworking files


odedle-tr:
### Description

Hi, we have an ios cordova app and we use a third party plugin that uses the AFNetowrking framework. 
We have the updated version of the AFNetwoking framework installed via pods. 

We also have your plugin installed wich uses the depnedancy **corodva-plugin-advanced-http.** 

The issue here is when installing your plugin, the dependency **corodva-plugin-advanced-http** overide the AFNetworking files, 
resulting in an older version of this framework. 

**We get the error in xcode resulting in an app crash:** 

(lldb) *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AFHTTPSessionManager GET:parameters:headers:progress:success:failure:]: unrecognized selector sent to instance 0x2817ed950'
terminating with uncaught exception of type NSException
(lldb)

**This happens because our third party plugin expect this inteface:** 

- (NSURLSessionDataTask *)GET:(NSString *)URLString
                   parameters:(nullable id)parameters
                      headers:(nullable NSDictionary <NSString *, NSString *> *)headers
                     progress:(nullable void (^)(NSProgress * _Nonnull))downloadProgress
                      success:(nullable void (^)(NSURLSessionDataTask * _Nonnull, id _Nullable))success
                      failure:(nullable void (^)(NSURLSessionDataTask * _Nullable, NSError * _Nonnull))failure

**But the corodva-plugin-advanced-http overrides it with probably an older version (see: 'headers' and 'progress' are missing in this api):**

- (NSURLSessionDataTask *)GET:(NSString *)URLString
                   parameters:(id)parameters
                      success:(void (^)(NSURLSessionDataTask *task, id responseObject))success
                      failure:(void (^)(NSURLSessionDataTask *task, NSError *error))failure

Please let know how we can overcome this issue or apply the correct fix 
Thanks.

### Additional Information

* cordova-plugin-code-push version:  1.12.0
* Cordova version: 9.0.0
* Cordova-ios versio: 6.2.0
* iOS/Android/Windows version: - 14.4
* Does this reproduce on a debug build or release build? - both.
* Does this reproduce on a simulator, or only on a physical device? - both

PR #682, opened 2021-07-27
Fix code push dependency "cordova-plugin-advanced-http" AFNewtworking overriding

meiram-tr:
when installing cordova code push plugin, it install a dependency plugin "cordova-plugin-advanced-http". The "cordova-plugin-advanced-http" is using a local copy of AFNetworking with manual changes. when this plugin is playing together with other dependency that depends on https://cocoapods.org/pods/AFNetworking version 4.0 and above this plugin overides the real AFNetworking library files. I have created a fork of the "cordova-plugin-advanced-http" and fix the issue.
In order to make both library and this plugin together I replaced AFNewtworking files and classes names (in this code) to prevent conflicts with the real pod library. else it causes the apps to crash.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -10,7 +10,7 @@
         <dependency id="cordova-plugin-dialogs" version=">=1.1.1" />
         <dependency id="cordova-plugin-device" version=">=1.1.0" />
         <dependency id="cordova-plugin-file" version=">=6.0.1" />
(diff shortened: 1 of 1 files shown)


meiram-tr:
The PR is fixing issue https://github.com/microsoft/cordova-plugin-code-push/issues/679

